PR #7, opened 2022-07-15
Implements a new configuration setting to ignore ASNs received from flow data

bcollet:
This new setting allows to ignore source and destination autonomous system numbers received from flow data and rely solely on MaxMind's data.

This is useful when an exporter only gets a partial routing table as well a default route from a transit provider. Without it all flows matched against the default route on the exporter have the transit provider's AS number instead of the origin's.

However, as I am not familiar with Akvorado's code, it might not be the best way to implements this and it might be interesting to have this setting available per exporter rather than globally.

### DIFF
--- a/console/data/docs/02-configuration.md
+++ b/console/data/docs/02-configuration.md
@@ -143,6 +143,9 @@ The following configuration keys are accepted:
   classifiers are pure, their result is cached in a cache. The metrics
   should tell if the cache is big enough. It should be set at least to
   twice the number of the most busy interfaces.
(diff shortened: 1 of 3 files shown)


vincentbernat:
Thanks!

vincentbernat:
I think this could be implemented as a list of priority:
```yaml
asn-source:
 - flow-except-private
 - geoip
 - flow
```
This would be helpful the day we have a third source (like BGP).
As doing that per-exporter, I prefer to wait for someone to have the need to implement this.

PR #23, opened 2022-07-18
inet/flow: add sflow support

tchadelicard:


vincentbernat (changes requested):
Also:
- add an entry in `docs/99-changelog.md` (at the top since this would be the most important feature of the next release)
- enable it in the configuration provided `akvorado.yaml`
- (optional) `cd docs; git grep -i netflow` and these parts may need updating
- (optional) you may want to `git rebase -i origin/main` to suppress the merge commits and maybe squash some of the commits

console/data/docs/00-intro.md
```diff
@@ -13,8 +13,8 @@ a web interface to browse the result.
 
 A `docker-compose.yml` file is provided to quickly get started. Once
 running, *Akvorado* web interface should be running on port 80 and an
-inlet accepting NetFlow available on port 2055. You need to configure
-SNMP on your exporters to accept requests from Akvorado.
+inlet accepting both NetFlow and sFlow is available on ports 2055/6343.
```
vincentbernat, line 16:
```suggestion
inlet accepting both NetFlow (port 2055) and sFlow (port 6343).
```

vincentbernat (changes requested):
Almost there!

PR #60, opened 2022-07-29
orchestrator/clickhouse: fix ORDER BY for summing merge trees

vincentbernat:
SrcCountry and DstCountry are missing from the ORDER BY clause of the
summing merge trees. Therefore, they are merged with a random value.
We try to fix that.

However, it is not possible to modify ORDER BY for existing columns.
We get the following error:

```
code: 36, message: Existing column SrcNetName is used in the
expression that was added to the sorting key. You can add expressions
that use only the newly added columns
```


vincentbernat:
Close in favor of #61.

PR #278, opened 2022-11-26
inlet/bmp: new lockless design

vincentbernat:
This reverts commit 7196ccf73b65c34960ce672d60b22b97f9cec82d, which was reverting 3199f8f1a79c245d53be5d7428d2e02c81fb4506 because this change may be buggy and I wanted to do a release.

I need to think if there are better ways to achieve the same goal (fast lookup first, low memory usage second) in a different way. The issue with locks is that they are fulfilled in the order they were taken. If we keep the lock design, readers should have priority over writer.


vincentbernat:
For this one, I am waiting for the current rewrite using an outlet component. This will help BMP batching updates.

PR #329, opened 2022-12-15
inlet/flow: allow ExporterAddress overwrite from flow source IP

margau:
This PR adds an option for selecting the ExporterAddress not from the Flow Message, but from the Source IP of the Flow Packet. The defaults are as usual.
This feature is helpful when the Software Generating flow messages sets the wrong Exporter Address for some reason.

### DIFF
--- a/console/data/docs/02-configuration.md
+++ b/console/data/docs/02-configuration.md
@@ -69,7 +69,8 @@ listening endpoint, `workers` to set the number of workers to listen
 to the socket, `receive-buffer` to set the size of the kernel's
 incoming buffer for each listening socket, and `queue-size` to define
 the number of messages to buffer inside each worker.
-With `agent-id-src-addr-overwrite` set to true, the source ip of the received flow packet is used as Exporter Address.
+With `use-src-addr-for-exporter-addr` set to true, 
+the source ip of the received flow packet is used as exporter address.
 
 For example:
 
--- a/console/data/docs/02-configuration.md
+++ b/console/data/docs/02-configuration.md
@@ -68,7 +68,10 @@ For the UDP input, the supported keys are `listen` to set the
 listening endpoint, `workers` to set the number of workers to listen
 to the socket, `receive-buffer` to set the size of the kernel's
 incoming buffer for each listening socket, and `queue-size` to define
-the number of messages to buffer inside each worker. For example:
+the number of messages to buffer inside each worker.
+With `agent-id-src-addr-overwrite` set to true, the source ip of the received flow packet is used as Exporter Address.
+
+For example:
 
 ```yaml
 flow:
@@ -77,10 +80,12 @@ flow:
       decoder: netflow
       listen: 0.0.0.0:2055
       workers: 3
+      agent-id-src-addr-overwrite: true
     - type: udp
       decoder: sflow
       listen: 0.0.0.0:6343
       workers: 3
+      agent-id-src-addr-overwrite: false
   workers: 2
 ```
 
--- a/console/data/docs/02-configuration.md
+++ b/console/data/docs/02-configuration.md
@@ -81,12 +81,12 @@ flow:
       decoder: netflow
       listen: 0.0.0.0:2055
       workers: 3
-      agent-id-src-addr-overwrite: true
+      use-src-addr-for-exporter-addr: true
     - type: udp
       decoder: sflow
       listen: 0.0.0.0:6343
       workers: 3
-      agent-id-src-addr-overwrite: false
+      use-src-addr-for-exporter-addr: false
   workers: 2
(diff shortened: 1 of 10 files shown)


console/data/docs/02-configuration.md
```diff
@@ -68,7 +68,10 @@ For the UDP input, the supported keys are `listen` to set the
 listening endpoint, `workers` to set the number of workers to listen
 to the socket, `receive-buffer` to set the size of the kernel's
 incoming buffer for each listening socket, and `queue-size` to define
-the number of messages to buffer inside each worker. For example:
+the number of messages to buffer inside each worker.
+With `agent-id-src-addr-overwrite` set to true, the source ip of the received flow packet is used as Exporter Address.
```
vincentbernat, line 72: Wrap to 80 columns.

console/data/docs/04-operations.md
```diff
@@ -7,6 +7,13 @@ service and accepts SNMP requests. For routers not listed below, have
 a look at the [configuration
 snippets](https://github.com/kentik/config-snippets/) from Kentik.
 
+### Agent ID/Exporter Address
+The Exporter Address is set from the field inside the flow message by default, and used e.g. for SNMP requests.
+However, if for some reasons the set flow address is wrong, you can use the source IP of the flow packet instead by setting `agent-id-src-addr-overwrite: true` for the flow configuration.
+
+Please note that with this configuration, your Deployment must not touch the Source IP!
+This might occur with Docker or Kubernetes Networking.
```
vincentbernat, line 15: Do not capitalize Exporter Address, Deployment, or Networking. Also, we don't use agent ID anywhere, so I think we should not introduce it just for this option. Maybe `use-src-addr-for-exporter-addr`?

inlet/flow/input/udp/root.go
```diff
@@ -185,6 +185,12 @@ func (in *Input) Start() (<-chan []*decoder.FlowMessage, error) {
 					Payload:      payload[:n],
 					Source:       source.IP,
 				})
+				// if the agent Id SrcAddrOverwrite is set, we actually overwrite the exporter Address
+				if in.config.AgentIDSrcAddrOverwrite {
+					for _, f := range flows {
+						f.ExporterAddress = source.IP
+					}
+				}
```
vincentbernat, line 193: This should be moved one level up, in the `inlet/flow` component. The `Decode()` methods also receive the source IP. This way, this would also work for PCAP files or for a future input method capturing from an interface.

vincentbernat (changes requested):
Also:
- use "inlet/flow: allow ExporterAddress overwrite from flow source IP" as the commit subject (if you are comfortable with git rebase, otherwise, I'll do it)
- add an entry in `console/data/docs/99-changelog.md`
- add a test if you can (otherwise, I'll do it)

inlet/flow/input/udp/root.go
margau: If i'm doing it in https://github.com/akvorado/akvorado/blob/28166569d26580ca8305a6b827aedb22a665d883/inlet/flow/root.go#L143 i don't have the respective input configuration any more. I'd keep the setting on a "per input configuration" however, to allow different settings for e.g. sflow/netflow/etc. - how do we solve this? We could try add the InputConfiguration to `type Input interface {`?

vincentbernat: I think you can do it here: https://github.com/akvorado/akvorado/blob/28166569d26580ca8305a6b827aedb22a665d883/inlet/flow/decoder.go#L27

margau: When doing it on that level, i'm having problems with the configuration at the moment:
For my understanding, this should go into 
```
// InputConfiguration represents the configuration for an input.
type InputConfiguration struct {
	// Decoder is the decoder to associate to the input.
	Decoder string
	// Config is the actual configuration of the input.
	Config input.Configuration
	// UseSrcAddrForExporterAddr replaces the exporter address by the UDP flow source.
	// default is false, meaning that the agent id inside the flowmessage is used
	UseSrcAddrForExporterAddr bool
}
``` 
to be generic for all input methods. I can access it in the decoder from there, however, it is not getting parsed correctly from the input configuration. How do I solve this?

vincentbernat: Keep it in UDP for now, I'll investigate this later (the solution is to also modify `ConfigurationUnmarshallerHook()` and `MarshalYAML()` functions, but I would like to do something more generic to avoid having to update these functions.

margau: Rolled back the change and tested it with the current setup, should work as it is now.

vincentbernat:
Thanks!

PR #369, opened 2022-12-27
Improve loglevel load

junnhy5:
Use a variable to store the global logging level rather than calling `zerolog.SetGlobalLevel` every time. This will reduce the number of function calls and improve code efficiency.


vincentbernat:
Sorry, this seems meaningless. You reduce the call to `zerolog.SetGlobalLevel()` from 2 to 1 (for the total program execution). Moreover, the function is so simple (it stores the provided value) that it is likely to be inlined directly by the compiler.

PR #384, opened 2023-01-03
Overriding flow data for 802.1q tagged traffic 

smallsam:
This is a work in progress of the functionality I need to update flow data based on the 802.1q vlan id tag to meet the requirements outlined in #318.

I wanted to get this work in progress on your radar as I was just about to work through the clickhouse schema changes but it looks like you're changing this up significantly with your recent commit and wanted to know if there was anything else coming that I need to be aware of:
https://github.com/akvorado/akvorado/commit/7d1ba478a1b8de5a52c0d8ab511e7fe95edddfa5

I will rebase on your latest develop branch next time I work on this but in the meantime any feedback you have on my current work would be much appreciated.

My current TODO includes:
1. Move regex compiles into config load, remove lock during classification.
2. Update clickhouse schema migrations


vincentbernat:
Your changes are likely easy to adapt to what I did, but don't bother now as more changes are coming.

I was expecting that you would move the VLAN handling inside the classifiers: adding VLAN ID to what can be queried by the interface and add more functions to fit your needs, including one to alter description. Is there something I am missing?

smallsam:
That was the original path but as I was experimenting with as per our initial thread and variations on that, including encoding vlan info in the ifAlias of the parent interface that but I realised the user experience was not good for the network admin and there was a risk that I'd exceed the char limit for an ifAlias with all that info.

When I realised that the cleanest path forward was to add the info statically (for now) into akvorado's config, I started to put them into a classifier. This became a little tricky, updating the interface classification cache key with the vlanid, then the environment for the classification program etc, and I didn't find a yaml/regex syntax that was easy to update for a network admin adding new peerings.

I wanted to reduce the risk of breaking the classification programs. So I landed on the yaml map, which is easy to edit and is clearer in both the config and code that the vlan tag only impacts the input interface.

vincentbernat:
Let's continue like this for now until you have something you are OK with. Don't spend too much time on the config part. I still prefer the expression language otherwise we will end up with people wanting to add something different, like adding the VLAN number only if it's different than 3, etc.

smallsam:
I was happy with the structure for now. I agree the expression felt cleanest but adding just the vlanid felt limiting so I wanted to generalise it to pass in the entire flow metadata as arbitrary dimensions for classification may be useful for others. However passing the entire flow metadata into the classification program effectively invalidates the entire interface classification cache and would need to run a classification program for every flow.

If you can let me know when your current changes are done and I'll rebase/fix up this branch to make it good.

If you have specific changes in mind that you'd like to see, I'd welcome your insight. This is the most go code I've written ever too, so any and all advice you have for be is awesome.

On your edge case you mention, I agree having it in the program makes these types of things possible, however I forsee users auto-generating these vlanid to description lists to import into the akvorado config from their NMS. (That would be our plan). Whilst not as elegant as a short go expression program, having a flattened precomputed list has a trivial benefit of removing any cache warming for interface classification.

vincentbernat:
Exporter classification needs exporter's related data to make a choice. Interface classification needs interface's related data. Adding country or IP address seems far stretched (and it would defeat the caching as you have noticed).

As for the configuration, we could have something like this as an expression:

```yaml
- |
    Interface.Description matches "MegaIX NNI" && (
     Interface.VLAN == 10 && SetDescription("IX: Amazon") ||
     Interface.VLAN == 20 && SetDescription("IX: Twitch") ||
     Interface.VLAN == 901 && SetDescription("Transit: Vocus") ||
     SetDescription("Customer: vlan " + Interface.VLAN))
```

The language is not the most expressive one and may be a bit odd, but it covers your needs and also many of the future needs.  You can also generate it from your NMS.

With static configuration, it sure fits exactly what you need, but I am pretty sure I will have people asking for something different.

I didn't finish the refactoring as I plan to also change the way protobuf is built. Feel free to keep the conflicts for now, I think they will be easy to solve once I am done.

smallsam:
That works.

I propose just one small change, adding a Flow.VLAN so as to distinguish from the interface SrcVlan and DstVlan on the interface (currently not exposed in akvorado).

Also just to double check, you intend that expression to be part of the list of interface-classifiers in the config?

vincentbernat:
Yes, the expression would be part of `interface-classifiers`.

As for `Flow.VLAN`, I don't quite understand. `Interface.VLAN` or `Flow.VLAN` would be set to the value of `SrcVlan`/`DstVlan` (depending if we are classifying input or output interface). If we were to expose the index, it would be `Interface.Index` while it would still comes from the flow. Same for exporters actually, we have `Exporter.IP` and it comes from the flow. It seems simpler to put everything in the `Interface` namespace.

smallsam:
The reason for this distinction is more clearly shown in:
```
	// Vlan
	SrcVlan uint32 `protobuf:"varint,33,opt,name=src_vlan,json=srcVlan,proto3" json:"src_vlan,omitempty"`
	DstVlan uint32 `protobuf:"varint,34,opt,name=dst_vlan,json=dstVlan,proto3" json:"dst_vlan,omitempty"`
	// 802.1q VLAN in sampled packet
	VlanId uint32 `protobuf:"varint,29,opt,name=vlan_id,json=vlanId,proto3" json:"vlan_id,omitempty"`
```
from: https://github.com/netsampler/goflow2/blob/main/pb/flow.pb.go#L115

https://github.com/netsampler/goflow2/blob/main/docs/protocols.md makes for interesting reading as it suggests that sflow has these as 3 distinct fields, whereas netflow 9 VlanId = SrcVlan and IPFIX the VlanId = DstVlan.

As I'd only been working with SFlow and switches that don't populate the extended switch info, I don't have SrcVlan or DstVlan populated.

I've only added support for the VlanId, which although gives different behaviour depending on protocol (TIL), still has something of interest.

On populating VlanId as Src/Dst depending on In/Out interface, it makes sense and was my original plan, until I saw that my sflow payload contained neither. My assumption up until today was to treat VlanId as aligned with the input interface but today's commit I've injected into both interfaces. However it's not clear what I should do in my situation where sflow only provides the vlanid and neither src/dstvlan are provided and I believe _could_ differ from both.

Perhaps the next decision is, do we need to propagate 2 vlan ids or is 1 sufficient, i.e. just the vlanid and make it interface agnostic (because it could be either the src or dst vlan)?

smallsam:
@vincentbernat for some reason, it's finally clicked for me. I was stuck with the distinction of the hierarchy in the sflow payload between the exporter info, the flow info and the packet header. I kept mapping these three into exporter, interface and flow in my mind. I will coalesce interface and flow now.

vincentbernat:
I didn't check myself the Src/DstVlan/VlanId problem, but this would mean a user cannot rely on just VlanId. I think you should pass to the `classifyInterface()` function the best you can do. Otherwise, the user may have incorrect information depending if the interface is being classified as input or output.

|SrcVlan|Source VLAN ID| |From ExtendedSwitch|SRC_VLAN (59)|vlanId (58)|
|DstVlan|Destination VLAN ID| |From ExtendedSwitch|DST_VLAN (59)|postVlanId (59)|
|VlanId|802.11q VLAN ID| |Included|SRC_VLAN (59)|postVlanId (59)|

It's a bit an odd choice for VlanId to have source for NFv9 and destination for IPFIX (maybe that's an error we could check in the code).

Assuming that you don't have extendedswitch because there is no pop/push of VLAN, I would assume that if SrcVlan == 0/DstVlan == 0/VlanId != 0, then, we can set SrcVlan = DstVlan = VlanId?

vincentbernat:
The build failure should be solved (I hope). Also, I hope you didn't build anything yet to run on your production, the aggregating views were incorrect. See https://github.com/akvorado/akvorado/commit/d54a099b08e256eea0a1408aa72fdaf44f4fed92.

vincentbernat:
It is still unclear for me if in your case SrcVlan = DstVlan = VlanId or not. That's what I would expect for a switch not handling extended switch attributes.

I did have a quick look at the changes and they look good. Maybe add a SetDescriptionRegex and SetNameRegex as this can be useful too and would be on par with the other functions.

vincentbernat:
Please, do not try to rebase your branch. There are other changes incoming. I'll ping you once this is done and I'll help you rebase.

vincentbernat:
The code base should be more stable from now on. Tell me if you want to modify your PR yourself or if you prefer I do that.

smallsam:
@vincentbernat thanks. I will make the updates to change out the single VlanId field in favour of having both Src/DstVlan fields and add in some suitable heuristics based on my observations of different switch attributes coming through.
If it's not clear to be how to handle the rebase I'll reach out, much appreciated!
In the meantime the team has been trialing an early build with VlanId to see if it meets their needs, so far so good and will build out production once this PR is in.

vincentbernat:
I have pushed ae1686abd6df that should fix this.

PR #588, opened 2023-03-23
common/schema: Add option to convert alias dimension to generated dimension

margau:
When using the Prefix functionality, we have problems with the relatively expensive alias-based query that is used currently, and for my understanding invoked on every query.

When the exact same conversion from IP and Mask into the full Prefix is done on consuming the flow, and the finished prefix is stored in the flows database, querying is much more performant. This also has the advantage that we might ship the prefix information into aggregation tables, and can provide more accuracy than just "DstAS" there.


margau:
I've adjusted the commit that the main table change is not included.
Do you have an idea on how to solve the alias vs. generated issue?
As said before, with our extensive usage of the Prefixes, Alias turned out to be too computationally expensive, so I'd go for the "more storage" path there.

vincentbernat:
I think the idea is to have an additional options of stuff to convert from alias to generate (like the option to not limit something to the main table). BTW, it should not be needed to specify DstNetPrefix, there should be something transforming it from SrcNetPrefix.

margau:
After Checking and trying it again, this PR is ready from our side.

vincentbernat:
Thanks!

PR #641, opened 2023-04-20
Update build docs to reflect necessary NodeJS version

rnavarro:
I tried to build the code using `make` with node `v14.16.0` but I got the following error:

```
failed to load config from /home/rnavarro/workspace/akvorado/console/frontend/vite.config.ts
error during build:
Error: Cannot find module 'node:path'

```

A little research showed:

`node:` protocol support for CJS was added only in v16.

After updating to v16+ running `make` succeeded

This PR updates the docs to reflect that.


vincentbernat:
Support was added in Node 14.18. But let's say we need Node 16 in the documentation.

PR #653, opened 2023-04-25
nix: proper reproducible build (WIP)

andreabedini:
Hi there,

This is WIP attempt to get akvorado to build properly inside nix. This approach aims to include all build dependencies and does not need `--impure` to build.

I basically had to reverse engineer the custom makefile and reproduce some of its logic. This is clearly not optimal, since developers now have two buildsystems to maintain. I am not familiar enough with the go builds system but I wonder if we could get nix to 'pre-fetch' the go dependencies and then just run the makefile.

The build itself succeds but the check phase fails, likely I am not building things correctly.

Let me know if there's interest in pursuing this. I'd love to be able to deploy akvorado automatically with NixOS :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ jobs:
     name: 🤖 Check dependabot status
     runs-on: ubuntu-latest
     permissions: {}
+    if: "!startsWith(github.event.head_commit.message, 'build: update flake.nix')"
     steps:
       - name: Fetch dependabot metadata
         if: ${{ github.actor == 'dependabot[bot]' && github.event_name == 'pull_request' }}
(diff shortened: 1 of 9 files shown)


vincentbernat:
Hello Andrea,

Sorry, I really don't want to maintain two build systems and I cannot impose Nix on everybody. I am not happy with `--impure` flag either, but I don't want to maintain synchronization between `go.mod` and `gomod2nix.toml`.

For build tools downloaded from the Makefile, you can just override the appropriate variable:

```sh
export GOIMPORTS=goimports
export PIGEON=pigeon
```

For everything with a hash, we should have a way to update them from the Makefile (or an helper script). This way, we run the Makefile for the impure part and we can use Nix for the pure part with the updated hashs. In the CI, we can even do a PR when there is a change.

andreabedini:
Hi @vincentbernat,

thank you for looking into this so quickly. I am totally on board with your POV. My make-fu is a bit rusty and I didn't think of using environment variables! I'll have another go :)

andreabedini:
Superseeded by #718

PR #718, opened 2023-05-18
build: tentative to build using Nix in pure mode

vincentbernat:
This currently does not work because of the mock package.

This PR reuse some of the code from #653. Two major changes:

- add a script to easily update hashs (in the future, this should include nixpkgs)
- do not duplicate the content of Makefile

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ jobs:
     name: 🤖 Check dependabot status
     runs-on: ubuntu-latest
     permissions: {}
+    if: "!startsWith(github.event.head_commit.message, 'build: update flake.nix')"
     steps:
       - name: Fetch dependabot metadata
         if: ${{ github.actor == 'dependabot[bot]' && github.event_name == 'pull_request' }}
--- a/.github/workflows/update-flake.yml
+++ b/.github/workflows/update-flake.yml
@@ -25,5 +25,5 @@ jobs:
         run: |
           git config --local user.name "github-actions[bot]"
           git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
-          git add flake.nix flake.lock
+          git add flake.lock nix
           ! git commit -m "build: update flake.nix" || git push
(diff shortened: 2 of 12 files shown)


vincentbernat:
The issue with mockgen is now fixed. It seems to work. The CI should trigger a PR when there is a change in go.sum or packages-lock.json.

vincentbernat:
@andreabedini what do you think of this PR?

I am not 100% sure to merge it or keep it because it makes everything more complex, notably, the action to keep `flake.nix` up-to-date.

vincentbernat:
About asn2org, it will be updated at the same time as the hash (it was part of `nix .#update`). Initially, I did modify the CI updating nixpkgs to also do a PR for that, but I didn't want too many PR and I was thinking that we'll get at least one commit a month and this commit will update asn2org as well. However, now the commit needs to modify `flake.lock` or `package-lock.json`, so maybe I can add back the few lines in `update-nix.yml`.

As for using overlays, it changes the default Go/node version for everything, even stuff we don't care. For example, it changes the Go version for pigeon, so it may trigger a recompilation while we are fine if pigeon does not use the Go version we use. That's why I didn't use an overlay.

As for passthru, OK.

As for how you compute hash, for Go, it means with an empty cache, go mod vendor will run twice. Once to compute the hash and once for buildGoModule. With the previous method, if the hash was already OK, we only run it once. I think the problem is similar with NPM. I don't like parsing stderr and I tried using nix-prefetch, but it does not work well with Flakes and internally it does the same thing.

vincentbernat:
If the hash is already correct, go mod vendor will only run once. This is "important" when building the docker image. To ensure the docker image always use the right hash, it refresh them before invoking nix. This is from a clear cache. If you run go mod vendor to compute/check the hash, it is run then a second time during build (since there is no nix cache).

As for using additional files, yes, it would be better.

vincentbernat:
Let me merge it to see how it looks like, we can iterate if needed.

vincentbernat:
First automated commit to update hashes: https://github.com/akvorado/akvorado/commit/8d6fcab03db66c439a395562b910eb97ab0e607d

PR #727, opened 2023-05-27
inlet: introduce metadata component with pluggable providers

vincentbernat:
SNMP is the first (and default) provider. Further commits should add:

- [x] SNMP coalescing (or at the metadata level?)
- [x] Configuration conversion
- [x] At least one other provider (static one?)


vincentbernat:
It would be useful if people interested in an HTTP source (from LibreNMS for example) to provide the HTTP request and response that could be used to get an exporter name from an IP address and interface name, description, speed from its index.

PR #732, opened 2023-06-01
feat: add bioris client as an alternative BMP solution

margau:
As we already have a RIS infrastructure based on Bio, we use those to enrich the flows with metadata instead of the internal BMP component.
This PR introduces a bioris component, which can be used as replacement for the internal bmp component.


margau:
@taktv6 @BarbarossaTM you might also wanna have a look?

inlet/bioris/config.go
```diff
@@ -0,0 +1,50 @@
...
+// ConfigurationUnmarshallerHook normalize BioRIS configuration
+func ConfigurationUnmarshallerHook() mapstructure.DecodeHookFunc {
+	return func(from, to reflect.Value) (interface{}, error) {
+		if from.Kind() != reflect.Map || from.IsNil() || to.Type() != reflect.TypeOf(Configuration{}) {
+			return from.Interface(), nil
+		}
+
+		return from.Interface(), nil
+	}
+}
```
vincentbernat, line 46: This can be removed.

margau: I've cleaned this up and added a rudimentary test based on the other inlet config tests.

vincentbernat:
Some preliminary remarks:
1. It would be great to have unittests
2. I would prefer to introduce a "ris" module with two implementations (no modification to inlet/core), like what was done in https://github.com/akvorado/akvorado/pull/727, but I don't mind doing that myself as this may be a bit of grunt work and a distraction to the main goal
3. I think that disconnections from the RIS servers are not handled (but maybe that's done transparently by the underlying gRPC lib?)

margau:
Regarding Unittests: I'll have a look and try to add some.

The GRPC Libary should already handle disconnects and automatically tries to reconnect, we already try to reconnect with a backoff. The LPM call itself has a Context Deadline of 200ms to not block the flow enrichment too long.

vincentbernat:
Ack. Ignore my remark about the additional module, we'll handle that later.

margau:
I've added unittests for the two more complex algorithmic parts (choosing the proper router/ris for LPM lookup, as well as converting the LPM result to the bmp LookupResult).

console/data/docs/02-configuration.md
```diff
@@ -112,7 +112,20 @@ Without configuration, *Akvorado* will listen for incoming
...
+
+There are two BMP/RIS implementations available:
+The default is the internal BMP component, intended to directly terminate BMP Sessions from routers.
+Alternatively, BioRIS is available, intended to lookup routing information from existing [bio-rd RIS](https://github.com/bio-routing/bio-rd/tree/master/cmd/ris) instances.
+
+You can choose by setting the `ris-provider`:
+```yaml
+inlet:
+  core:
+    ris-provider: bioris
```
BarbarossaTM, line 125: How would you folks feel about using `BioRIS` here as is used internally?

BarbarossaTM (commented):
This is neat, thanks for your efforts! <3

console/data/docs/02-configuration.md
margau: I'd say the current, lower case bioris variant fits best to the existing scheme: https://github.com/akvorado/akvorado/blob/main/inlet/core/config.go#L72

console/data/docs/02-configuration.md
```diff
@@ -141,6 +154,27 @@ space used in ClickHouse.
 *Akvorado* supports receiving the AdjRIB-in, with or without
 filtering. It may also work with a LocRIB.
 
+#### BioRIS component
+
+The connection to an existing BioRIS instance is configured with the following keys:
+```yaml
+inlet:
+  bioris:
```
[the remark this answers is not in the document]
margau: See above comment

vincentbernat:
I am currently a bit busy, but I'll work on it before end of month.

vincentbernat:
I am unable to push changes to your branch. Did you disallow modifications when opening the PR? I have pushed some small cosmetic fixes in https://github.com/akvorado/akvorado/tree/feature/bioris (this is also rebased on the current main).

As for testing, as BioRIS is also written in Go, I think it should be possible to run some tests against the real server, instead of using mocked responses for GRPC. Otherwise, I fear this can break unexpectedly. This can be done later. BTW, how stable is the GRPC interface? @BarbarossaTM 

What would happen if a RIS instance gets a new router? I think the list of routers should be refreshed periodically. This would also allow you to try to continue to open unreachable instance in the background if needed.

margau:
Thanks for the feedback!

Allow edits for maintainers seems to be broken here, as this is an org (and not personal) fork - i rebased it instead.

Unfortunately, the new router/new VRF case (as well as old one disappearing) is not yet properly solved beyond polling in upstream Bio, I have this issue also with the Bio RIS Mirror. 
I'm currently thinking about a Streaming RPC call announcing changes in Router/VRF availability, with that we could implement proper router/VRF changes at runtime. If this is a blocker, we could try and workaround this with polling the GetRouters RPC call in the meantime.

Regarding the random router: Yes, we could remove this, I'd just like to keep a trace in the metrics that we did not found a matching router.

What is the current plan with this going forward, waiting for the modular bmp support and rebase ontop of this, or get it ready now and do changes with the modular BMP?

vincentbernat:
Not having an up-to-date information is not a blocker to merge.

For the next steps, as you want. You can continue doing modifications: as long as they are only in `inlet/bioris` it should be very simple to merge. Or you can wait for the modular support that should be ready this week or the next.

margau:
The router selection topic should be addressed now, I also enhanced the documentation a little bit.

From my side, there are no more bioris modifications that should be part of this MR as of now.

vincentbernat:
See https://github.com/akvorado/akvorado/pull/802. This is not fully complete (it misses the configuration conversion from old format and I didn't really test for real).

I think this should be "easy" to convert the bioris component to a provider:
1. throw away any modification outside `inlet/bioris`
2. move `inlet/bioris` into `inlet/routing/provider/bioris` and turn the component into a provider (`New()` is a method from `Configuration`)
3. register the provider in `inlet/routing/config.go`

I think I can do that myself if you prefer.

margau:
I'll move this to a new MR against #802 in the next days. Still have some open points, but lets do this over in #802.

vincentbernat:
Let's close this one in favor of #809. I'll merge #802 this week-end.

PR #802, opened 2023-07-16
inlet/routing: introduce routing component with pluggable providers

vincentbernat:
BMP is the first and default provider.


vincentbernat:
@margau I have added some tests by spawning a BMP receiver and a RIS server. However, it is a bit odd as it seems that the routers I get are only "127.0.0.1", while I was expecting 2001:db8::3, 2001:db8::7, 192.0.2.1 and 192.0.2.5, like here: https://github.com/akvorado/akvorado/blob/main/inlet/bmp/root_test.go#L196. Am I doing something wrong?

cc @taktv6 @BarbarossaTM

PR #809, opened 2023-07-19
inlet/routing: add bioris routing information provider

margau:
This PR replaces #732, and depends on #802 

It adds the bioris client as alternative routing information provider next to the default, local BMP one.

Some changes are made to the interface:
- the lookup now also gets the exporter address, which enables matching of the correct rib
- errors are handled by the "parent" routing component

Local testing of the change is still to be done.


inlet/routing/root.go
```diff
@@ -71,6 +74,12 @@ type stopper interface {
...
 // error, even when the context times out. Instead, it should just returns an
 // empty answer.
-func (c *Component) Lookup(ctx context.Context, ip netip.Addr, nh netip.Addr) provider.LookupResult {
-	return c.provider.Lookup(ctx, ip, nh)
+func (c *Component) Lookup(ctx context.Context, ip netip.Addr, nh netip.Addr, agent netip.Addr) provider.LookupResult {
+	c.metrics.routingLookups.WithLabelValues().Inc()
+	result, err := c.provider.Lookup(ctx, ip, nh, agent)
+	if err != nil {
+		c.metrics.routingLookupsFailed.WithLabelValues().Inc()
+		c.r.Warn().Err(err).Msgf("routing: error while looking up %s at %s", ip.String(), agent.String())
```
vincentbernat, line 82: Only log rare occurrences or use a rate limiter (check `errLoger` in `inlet/metadata/provider/snmp/root.go`)

Andre4s: > Only log rare occurrences or use a rate limiter (check `errLoger` in `inlet/metadata/provider/snmp/root.go`)

Man thx for akvorado! But now i get message 404 page not found. After remove demo.yaml. what should I do?

vincentbernat:
Just to be sure: I'll wait for your ping for when this PR is ready!

inlet/routing/root.go
margau: Implemented with a rate limiter

margau:
> Just to be sure: I'll wait for your ping for when this PR is ready!

Should be ready now. I tested this with bioris locally, seems to work fine.

margau:
Quick feedback: This works good for us, thanks!

Issue #832, opened 2023-08-07
`docker-compose.yml` broken: Bitnami Kafka container now using KRaft by default


MattKobayashi:
```
# akvorado version
akvorado v1.8.3
  Build date: 2023-04-28T17:13:46+0000
  Built with: go1.20.3
```

# Bug description

The upstream Bitnami Kafka container has been updated and now appears to use KRaft by default. It no longer recognises the `ALLOW_PLAINTEXT_LISTENER` or `KAFKA_ENABLE_KRAFT` environment variables as defined in the repo's `docker-compose.yml` file. See here for details: https://github.com/bitnami/containers/pull/42964/files

## Steps to reproduce the problem

Start a fresh Akvorado instance from the repo's `docker-compose.yml` file.

## Expected outcome

For Akvorado to run successfully.

## Current outcome

Kafka fails to start due to the following error: `java.lang.IllegalArgumentException: Error creating broker listeners from 'PLAINTEXT://:9092': No security protocol defined for listener PLAINTEXT`. This appears to be related to the upstream changes mentioned above.

MattKobayashi:
## Temporary Workaround

If you're reading this issue because your Akvorado instance has stopped working, you can resolve this issue temporarily by pinning a specific version of the Bitnami `kafka` container. In `docker-compose.yml`:

```
  kafka:
-    image: bitnami/kafka:3.3
+    image: bitnami/kafka:3.3.2-debian-11-r175
    environment:
```

vincentbernat:
This is a bit annoying that Bitnami does not provide "stable" images. See the list of [notable changes](https://github.com/bitnami/containers/blob/main/bitnami/kafka/README.md#notable-changes).

vincentbernat:
This should be fixed by https://github.com/akvorado/akvorado/commit/79de41ad3421444548756f3053b147e56c281b4f.

PR #834, opened 2023-08-07
Pin `kafka` container image to specific release

MattKobayashi:
- Temporary workaround for #832


vincentbernat:
Thanks. However, I don't want to not get security updates. These kind of "silent" updates from Bitnami are a bit annoying and I wish they provided more stability.

vincentbernat:
Well, in the meantime...

Issue #759, opened 2023-06-28
gNMI for flow enrichment


hellt:
Hi @vincentbernat and the rest of the Akvorado team,

It is 2023 and maybe it is time to make a beautiful Akvorado even more awesome by adding modern data retrieval protocols such as Openconfig's [gNMI](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md)?

This will help bring more DC fabrics to enjoy Akvorado, as some of them started to switch to gNMI from SNMP.

We can use [openconfig/gnmic](https://gnmic.openconfig.net/) API to integrate it nicely into the Go codebase of Akvorado, and it should be rather a straightforward thing to do.

Looking forward to hear your thoughts on that, collaborate and bringing the future closer to Akvorado enrichers.

/cc @karimra @jbemmel

protonjhow:
To add some colour here, I enjoyed Will's talk at SwiNOG, and was playing around with Akvorado in our Lab. It's certainly more approachable than our goflow/pipeline/clickhouse adventures to date. For that, I am hugely appreciative of the work to date.

At Proton we require "secured" transport for telemetry and signalling wherever possible and whilst Akvorado supports SNMPv3, Nokia SR-L for example, does not. In fairness, the gNMI telemetry they offer is miles better so I don't miss SNMP... except for edge cases like this.

My Go skills are minimal, but I am happy to chip in if I can, obviously test and provide usable feedback to smarter people than me to move this forward as well. ;)

vincentbernat:
The SNMP part was made modular last month, so it's easier now than it was before.

vincentbernat:
First question to implement that: should we use "Get", "Subscribe" or have a configuration on the exporter side on what to push?

hellt:
You can use a simple Get given that retrieved amount data is not huge

vincentbernat:
A few followup questions:
- does the library handles reconnect automatically?
- should encoding be user configurable? what the default should be?

karimra:
> * does the library handles reconnect automatically?

Yes.

> * should encoding be user configurable? what the default should be?

Yes, not all routers support the same set of encodings, the default per the gNMI spec is JSON (ENUM value=0)

There is a Get RPC example [here](https://github.com/openconfig/gnmic/blob/main/examples/pkg/get_rpc/main.go)

jbemmel:
Encoding could be automatically selected based on capability discovery
response

vincentbernat:
[...]

PR #840, opened 2023-08-08
feat: add custom dictionaries for additional, customized flow hydration

margau:
In our network, we have the need for additional metadata attached to flows, which cannot be covered by the Classify-Logic, but the clickhouse dictionaries are a perfect fit for this.

This PR adds the capability to use a custom dictionary for custom dimensions with flowhouse.
Clickhouse is configured to do the usual dict lookup, and fetch the necessary CSV from the orchestrator. This is "materialized", meaning the dict lookup happens at ingest time (and not at query time), as this leads to more data consistency with changing custom data.

With this PR, the csv is loaded from a file, but I'm planning to also do an S3 option.

One point is still open, where I'm looking for input: Currently, it is not possible to filter for these new dimensions, as they are added at runtime, but the filter grammar is build at compile time.
There is probably a way to fetch the key names in the parser during runtime, however I did not found it yet.


vincentbernat:
I am OK with such a feature. Adding tests would be useful. I need to work a bit on #759 before looking at this one.

margau:
A first part of the tests is pushed.
The migrations are a WIP, there is an issue with the "MODIFY ORDER BY"-columns for the downsampled tables (coming from https://github.com/akvorado/akvorado/blob/92506280b80adebc9e7b733c74e957b9f554bf6c/orchestrator/clickhouse/migrations_helpers.go#L497), need to have another look there

vincentbernat:
I was a bit confused by the configuration part first. After reading it again, I think `dimensions` apply to both keys (where they can be overridden with `match` and `match-post`) and values (where the can be completed with `label` whose default value is a capitalized version of `name`).  I would propose to rename `match` to `match-dimension` and `match-post` to `match-dimension-suffix`. I am OK with `label`.

common/schema/config.go
```diff
@@ -17,6 +17,27 @@ type Configuration struct {
...
+	// CustomDictionaries allows enrichment of flows with custom metadata
+	CustomDictionaries map[string]CustomDict `validate:"omitempty,dive"`
+}
+
+// CustomDict represents a single custom dictionary
+type CustomDict struct {
+	Keys       []CustomDictElement `validate:"dive"`
+	Attributes []CustomDictElement `validate:"dive"`
+	Source     string
+	Layout     string `default:"hashed"`
```
vincentbernat, line 29: This is not a recognized metadata keyword. Have a look at `orchestrator/clickhouse/config.go` for `DefaultNetworkSourceConfiguration()` and its use in `init()`.

common/schema/config.go
```diff
@@ -17,6 +17,27 @@ type Configuration struct {
...
 	Materialize []ColumnKey
+	// CustomDictionaries allows enrichment of flows with custom metadata
+	CustomDictionaries map[string]CustomDict `validate:"omitempty,dive"`
+}
+
+// CustomDict represents a single custom dictionary
+type CustomDict struct {
+	Keys       []CustomDictElement `validate:"dive"`
+	Attributes []CustomDictElement `validate:"dive"`
+	Source     string
```
vincentbernat, line 28: A source is mandatory, so `validate:"required"`.

common/schema/config.go
vincentbernat: And add `validate:"required,oneof=hashed iptrie complex_key_hashed"` (not 100% sure, but I think this is better to support only some sane choices first).

common/schema/config.go
```diff
@@ -17,6 +17,27 @@ type Configuration struct {
...
+}
+
+// CustomDictElement represents a single column of a custom dictionary
+type CustomDictElement struct {
+	Name      string
+	Type      string
+	Label     string
+	Match     string
+	MatchPost string
+	Default   string
```
vincentbernat, line 40: Add some validation. `validate:"required,alphanum"` for  `Name`, `validate:"required,oneof=String UInt8 UInt16 UInt32 UInt64 IPv4 IPv6"` for `Type` (if we need to match, we need to not allow anything wild here). `validate:"alphanum"` for `Label`, `Match` and `MatchPost`.

Moreover, from my understanding, `Label` and `Default` are for values and `Match` and `MatchPost` for keys. I would prefer to have two seperate types for that (`CustomDictKey` and `CustomDictValue`).

orchestrator/clickhouse/http.go
```diff
@@ -90,6 +93,29 @@ func (c *Component) registerHTTPHandlers() error {
 			w.Write(result.Bytes())
 		}))
 
+	// add handler for custom dicts
+	for k, v := range c.d.Schema.GetCustomDictConfig() {
+		// we need to call this a func to avoid issues with the for loop
+		func(k string, v schema.CustomDict) {
+			c.d.HTTP.AddHandler(fmt.Sprintf("/api/v0/orchestrator/clickhouse/custom_dict_%s.csv", k), http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
vincentbernat, line 100: Another method is to do `k := k`, `v := v`.

common/schema/definition.go
```diff
@@ -377,10 +378,10 @@ END`,
 					`concat(toString(ICMPv6Type), '/', toString(ICMPv6Code))), '')`,
 			},
 		},
-	}.finalize()
+	}.finalize(ColumnLast)
 }
 
-func (schema Schema) finalize() Schema {
+func (schema Schema) finalize(lastKey ColumnKey) Schema {
```
vincentbernat, line 384: Isn't the information included in the schema?

margau: ColumnLast is a const variable pointing to the last hardcoded dimension. With the "flexible", custom-dict based dimensions, the last column identifiert needs to move, thats why I made it flexible with that argument.

Shall we move the actual last key into the schema struct, or use another option?

common/schema/config.go
margau: Improved with 542f36cb3cb5c771d1e408eeb8c9360806f7b7a2

common/schema/config.go
margau: Improved with 542f36cb3cb5c771d1e408eeb8c9360806f7b7a2

common/schema/definition.go
vincentbernat: `ColumnLast` was a very small optimization in `finalize()`. Instead of modifying a lot of call sites, just keep the current signature. You could replace `ColumnLast` by `len(schema.columns)` or just use `append()` to update `schema.ColumnIndex`.

margau: Taking only the len does not work due to disabled columns, and I'd try not to mess with the index outside finalize.
I've added the schema.dynamicColumns variable, clearly indicating how many dynamic columns are added after the ColumnLast.

margau: With `len(schema.columns)` in common/schema/definition.go:507, I've got 
```
 runtime error: index out of range
```
 
What implementation do you mean exactly with append? Directly appending to schema.columnIndex from the column creating loop in schema/root.go?
 
Another idea would be to provide a `schema.appendColumn(Column)` (or similar) method, encapsulating everything that is needed to add a new column to an existing schema.

How shall we continue with this inside this PR?

vincentbernat: Yes, instead of building the array at the right size from the beginning, use an empty one and append to it.

margau: Where would you set/append the columnIndex list in this case? If I append inside the New method in root.go, it would be overwritten by the last finalize in https://github.com/akvorado/akvorado/pull/840/files#diff-08a5918f6f118e79e5971a4585ae6ea800df9e6d7527eb29bdba1ca5ce3dc623R147 as far as I can tell.

vincentbernat: I lured you in the wrong direction as I didn't realize this was an inverted index. Instead, I did that: d1cef41849e7.

PR #850, opened 2023-08-25
feat: expose nexthop as dimension

margau:
This PR exposes the prepared NextHop dimension fully to database & frontend.

Nexthop is taken either from the flow or the routing component, I've reused  the NetProvider config for this, as a custom option only for the nexthop is possibly overkill.


common/schema/definition.go
```diff
@@ -376,6 +378,11 @@ END`,
 					`dictGetOrDefault('icmp', 'name', tuple(Proto, ICMPv6Type, ICMPv6Code), ` +
 					`concat(toString(ICMPv6Type), '/', toString(ICMPv6Code))), '')`,
 			},
+			{
+				Key:             ColumnNextHop,
+				ClickHouseType:  "IPv6",
```
vincentbernat, line 383: I think there are not many nexthops available. You could go with `LowCardinality(IPv6)` instead.

vincentbernat:
Looks good. You can also add an entry to the changelog?

vincentbernat:
I'll do the minor fix to not bother you!

vincentbernat:
I have disabled this new column by default. New columns should be opt-in as they can eat some space. We may change that if there is a benefit of having the column for everybody and after looking at how much space the column uses on a typical setup compared to the other non-default columns.

Issue #852, opened 2023-08-28
console: support filtering for custom dictionaries


margau:
With https://github.com/akvorado/akvorado/pull/840, I've added the option for custom dictionaries, creating custom dimensions. 

They are not supported in query filters (yet), which is what I wanna discuss here.

From my understanding the filtering is largely done in https://github.com/akvorado/akvorado/blob/main/console/filter/parser.peg, with the dimensions known at compile time.
As custom dict dimensions are done inside the configuration, they are only known during runtime.

I've not found a good way of doing this yet, are there ideas how we could dynamically add the custom dictionary dimensions to the parsing grammar?

vincentbernat:
This should be possible with the PEG grammar. There is a limitation for `c.metaColumn()` as it is not able to get the name. See https://github.com/mna/pigeon/issues/105. Maybe there is another way. As for `c.acceptColumn()`, it can and it could accept any column matching a real column of the right type.

vincentbernat:
See 9106a09bf1bc. You should be able to be more generic by having `c.acceptColumnInteger()`, `c.acceptColumnString()`, and stuff like that.

vincentbernat:
I have pushed 2f8fdc479afe which is an improvement over 9106a09bf1bc8859784c3d3358e19de89ba46550. I didn't test anything else yet, but the idea is that instead of having:
```
ConditionUintExpr "condition on integer" ←
 column:("InIfSpeed"i !IdentStart { return c.acceptColumn() }
       / "OutIfSpeed"i !IdentStart { return c.acceptColumn() }
       / "SrcPort"i !IdentStart { return c.acceptColumn() }
       / "DstPort"i !IdentStart { return c.acceptColumn() }
       / "SrcPortNAT"i !IdentStart { return c.acceptColumn() }
       / "DstPortNAT"i !IdentStart { return c.acceptColumn() }
       / "SrcVlan"i !IdentStart { return c.acceptColumn() }
       / "DstVlan"i !IdentStart { return c.acceptColumn() }
       / "PacketSize"i !IdentStart { return c.acceptColumn() }
       / "ForwardingStatus"i !IdentStart { return c.acceptColumn() }
       / "IPTTL"i !IdentStart { return c.acceptColumn() }
       / "IPTos"i !IdentStart { return c.acceptColumn() }
       / "IPFragmentID"i !IdentStart { return c.acceptColumn() }
       / "IPFragmentOffset"i !IdentStart { return c.acceptColumn() }
[...]

PR #870, opened 2023-09-07
feat: enable filters for custom dimensions

margau:
This MR is a first implementation step for https://github.com/akvorado/akvorado/issues/852

With this MR, there can now be filters for runtime-defined dynamic Dimensions (e.g. from custom dicts). 
This works by adding a "catch all dimensions" Expression to the filtering grammar, which explicitly excludes well-known dimension names (thats why there is a large diff in parser.peg, separating column name literals from expressions). 

For now, it is assumed that all custom dimensions are of type string or compatible.

Completion does not yet work for the custom dictionaries, the full dimensions needs to be entered manually.

According to the unit tests and my manual tests, the functionality for other dimensions should have been kept.

### DIFF
--- a/console/data/docs/02-configuration.md
+++ b/console/data/docs/02-configuration.md
@@ -559,8 +559,8 @@ to enrich your flow with additional informations not possible to get in the
 classifier. This works by providing the database with a CSV files containing the
 values.
 
-**Note:** filtering by dictionaries is not possible with the current state of
-development.
+**Note:** filtering by dictionaries is at the moment possible 
+with string-based custom dimensions.
 
 ```yaml
 schema:
--- a/console/filter.go
+++ b/console/filter.go
@@ -12,6 +12,7 @@ import (
 	"github.com/gin-gonic/gin"
 
 	"akvorado/common/helpers"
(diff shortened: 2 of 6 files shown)


margau:
I've also added the completion part as separate commit (if feasible, I could also split this up into separate PRs).
autocompletion is done by directly looking into the schema, as the grammar is not able to provide the column names as literals.

vincentbernat:
I was thinking it would be possible to be smarter in `filter.peg`. The directions I provided to you were not totally correct but I just pushed a commit showing how to do that. I have reused some of your tests to check if everything was working as expected.

What's missing:
- IPv4 handling
- value completion (you already have the code, I didn't want to "steal" more than what's necessary)

Maybe it would be easier to have another PR for that?

vincentbernat:
Done in 6e18065be663.

vincentbernat:
Note that I did remove the ability to use IPv4 as a type. People should use IPv6 (even if they only have IPv4) as all the code is handling IPv6, not IPv4. It wasn't even possible to display IPv4, I think.

PR #872, opened 2023-09-08
feat: add option for materialized types & improve filter performance for materialized prefixes

margau:
This PR adds some improvements to materialized dimensions, especially prefixes:
- it saves the Src/DstNetPrefix as LowCardinality, as we have a high repeatability here
- a flag indicates if a column was materialized
- the filters can be adjusted to use a more performant direct query in case of materialized columns, which is done for prefixes
- tests for the "fast" materialized prefix query are added


vincentbernat:
Thanks!

Issue #605, opened 2023-04-02
Support for ReplicatedMergeTree in ClickHouse


vincentbernat:
Currently, tables are created as `MergeTree`, preventing the ability to use a clustered ClickHouse. To support `ReplicatedMergeTree`, we should:

- Accept several ClickHouse servers (both console and orchestrator)
- Switch to `ReplicatedMergeTree` when we have several ClickHouse servers
- Execute `CREATE`/`DROP` on all members of the cluster
- Execute `ALTER` only on one member

Optionally:

- Configure ClickHouse with a Zookeeper cluster
- Convert the existing `MergeTree` tables to `ReplicatedMergeTree` (this seems safe, see the [documentation](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replication#converting-from-mergetree-to-replicatedmergetree), there is little point of going the other direction)

drMARTINS13:
I would also appreciate to run akvorado in some cluster mode.
It is not problem to prepare clickhouse and zookeeper cluster on another machine, but without support in akvorado it is useless.
Thank you

netixx:
@vincentbernat any reason you mention that ALTER need to be done only on one member ?

It looks to me like this is covered by using "ON CLUSTER": https://clickhouse.com/docs/en/sql-reference/distributed-ddl

We were also looking at supporting clickhouse clustering in our setup, hopefully we can make a contribution in the next few weeks.

I think we would like to go one step further and support sharding as well (through distributed table): https://clickhouse.com/docs/en/engines/table-engines/special/distributed. Making it optional seems like a must because it requires an heavier setup on clickhouse side. But it brings the bonus on not relying on the size of one individual server for max scale.

vincentbernat:
I don't remember. Maybe I misread the documentation or maybe ON CLUSTER was not available for CREATE in the past.

vincentbernat:
I'll work on that next as we now need it too.

PR #891, opened 2023-09-25
feat: create helm template

saymonaraujo:
This is the first commit that creates a helm template for deployment of Akvorado.

Todo:

Add probes
Improve README(add all variables)

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -1,0 +1,8 @@
+apiVersion: v2
+name: akvorado
+description: flow collector, enricher and visualizer.
+
+type: application
+version: 1.9.0
+appVersion: "1.9.0"
+icon: https://avatars.githubusercontent.com/u/113047722?s=48&v=4
--- a/helm/README.md
+++ b/helm/README.md
@@ -1,0 +1,98 @@
+# Akvorado Helm Template
+
+This is a simple Helm template for setting up your Akvorado namespace with the IPInfo container within your Google Kubernetes Engine (GKE) cluster.
+
+**Note:**
+- Compatibility has been tested only with GKE.
+- The shared PV from the IPInfo container has not been tested.
+
+## Features
+
+Out of the box, this template provides the following features:
+
+- Metrics collection for Akvorado within Prometheus using service discovery.
+- A Grafana dashboard for visualizing GKE Pods and Akvorado Metrics.
+
+## Passing Secrets
+
+You can pass secrets directly into the Helm template by placing them within the `secrets` section in the specific configuration. For `orchestrator`, use `akvorado_orchestrator`, and for `ipinfo`, place it within `ipinfo`. This allows you to use environment variables to change the Akvorado configuration, or you can update the configuration files inside the [config directory](config/).
+
+You can also use a specific image of Akvorado if you decide to run your own image and tag.
+
+## Akvorado Configuration
+
+Configuration of Akvorado itself should be done. Here are some of the parameters you should configure:
+
+- `config/akvorado.yml`:
+  - `kafka: brokers`
+  - `clickhouse: orchestrator_url`
+  - `clickhouse: servers`
+  - `username`
+  - `password`
+  - `database`
+  - `asns`
+
+- `config/console.yml`: If you want to use Redis for caching, you should configure Redis in this section.
+
+- `config/inlet.yml`:
+  - `snmp: communities`
+  - `classifiers`
+
+## Resource Limits
+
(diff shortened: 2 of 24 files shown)


vincentbernat:
As I am not familiar with Kube, I need some time for this.

The overall sentiment is that this is a lot of code that I would need to support forever. Ideally, it would have been nice that either the docker-compose setup could be translated to Helm or the other way to avoid duplication. Notably, if I need to update the docker-compose part, I need to be able to also update the Helm chart.

In the meantime, people are welcome to get your work through this PR.

saymonaraujo:
I still have a lot of work to do here. The current implementation for the shared PV/PVC for the ipinfo data does not work out of the box, instead of using the current approach I will change to use a NFS endpoint.
Another thing that I will add is the auto rollout of the pods if a change in the configmap is detected, this will provide a 'dyamic' configuration change(great if you use CI for manage the deployments/config).
I think that would be great to have a deployment of the ClickHouse and Kafka out of the box as well(still evaluating this).
There's some fix as well that I will commit once I finishes some clean up.

margau:
Hi,
we're also experimenting with akvorado and helm, thats an interesting PR!

Some thoughts after a first look:
- I'm not sure if it is possible to do the "config map change - restart deployments" with helm values only. The process I have in mind:
  - config map changes
  - orchestrator is restarted
  - if the orchestrator succeeded to load the config, it triggers a new deployment of the ingest and console (important: RollingUpdate to not loose flows). This might be done by annotating the pods with a hash of the config?
  - if the newly started orchestrator fails loading the config, the pods continue to run on the old config (no restart triggered)
  - if both "consumer" pods (inlet/console) and the orchestrator are restarted simultaneously, there is the risk of the consumer pods being ready first and loading the old config
- Monitoring: I'd say ServiceMonitor-resources might be a good option
- Storage: I'd try to keep inlets stateless (without PV), that way they can be scaled without issues, which is good for multi-node clusters with high input rates
- Orchestrator: I'm a bit worried about the number of orchestrators running during an update, not sure what the behavior with more than 1 orchestrator (e.g. the new one failing due to bad config) is
- Clickhouse: We're experimenting with altinity clickhouse operator, this might even integrate good with #605 in the future

saymonaraujo:
In my pocs the strimzi kafka operator and the altinity clickhouse operator worked really well. 
About the rollout after a confimap I will do a poc with the (reloader)[https://github.com/stakater/Reloader].

saymonaraujo:
I created a NFS Server that is used to share the files between the inlet and the ipinfo containers. I'm also in the final tests of the reloader. I'm adding a clickchouse and a Kafka in the helm template in the next commits.

saymonaraujo:
The reloader is working very well for Orchestrator and Console. But I'm having a little problem with the rollout of the inlet.
I'm working on this and soon will commit the new changes with the NFS server and the reloader.

netixx:
Stumbling upon that,

with https://github.com/akvorado/akvorado/pull/1059 it removes the need to have ipinfo/geoip data shared between the inlet, as only the orchestrator requires it.

tobikris:
I have thought about several ways to tackle the rollout of configuration changes in K8s.
I came up with the following options:

1. rollout inlet in orchestrator sidecar after orchestrator container started up
2. inlet init container that waits for orchestrator readiness (how to check that it is the updated orchestrator?)
3. rollout inlet in helm post-upgrade hook
4. load config in inlet on schedule (needs hot-reload in inlet)
5. load configmap in inlet directly without relying on orchestrator.

Option 5 seems the most straightforward solution, but I am wondering if this has downsides I am not seeing. It could happen that the updated inlet is trying to send data with new dimensions that have not been configured by the orchestrator - however it cannot be ensured that everything is strictly compatible at all times during upgrades anyway. If the orchestrator is updated first, the old inlet config can also lead to a now incorrect schema. I guess this is an expected situation during upgrades? What we only need to ensure is that all components are eventually using the new config. Or is this assumption incorrect?

vincentbernat:
Each schema uses a different Kafka topic, so the order should not be an issue.

tobikris:
> Each schema uses a different Kafka topic, so the order should not be an issue.

Ah, right! Glad I have asked. And about the config content itself - do you think it causes issues to not load the config from orchestrator? I did not find any hint of "magic" done by orchestrator and think it is just serving the plain file for convenience? Or is there anything else to consider?

vincentbernat:
No, it should be OK. It transforms a bit the configuration for migration, but inter-service configuration migrations should be rare.

PR #956, opened 2023-11-17
feat: add decoded flows/queue ingest metric for inlet/udp

margau:
This PR adds a metric counting the number of flows sent by the UDP input into the internal processing queue.

With the help of this metric, it is possible to determine the current size of the internal processing queue when debugging delays/dropped flows:
```
akvorado_inlet_flow_input_udp_decoded_flows_total - akvorado_inlet_core_received_flows_total
```


vincentbernat:
Thanks!

PR #960, opened 2023-11-22
Mpls label feature

ailtchev:
Storing MPLS fields if present in netflow/ipfix samples


ailtchev:
Hello,

I also thought about it, storing it like this will be more efficient. No
problem for me at all to let you go on from, I am interested to see how
pcap based unitary tests look like.

Le mer. 22 nov. 2023, 8:56 PM, Vincent Bernat ***@***.***> a
écrit :

vincentbernat:
I have opened #973 to store MPLS labels inside an array with access to 1st, 2nd, 3rd labels, in addition to be able to look for any label present in the stack. There is also an untested support for sFlow (parsing was already present).

I didn't create a group as the check for Netflow is quite simple, and parsing is done in all cases in sFlow.

There is a test using your pcap if you are curious. If you happen to be able to do a capture of a sFlow packet too with MPLS labels, it would also be useful!

PR #981, opened 2023-11-28
feat: add support for S3 as external data source

margau:
This PR adds support for S3:
- the orchestrator is a central S3 proxy
- currently, loading custom dictionaries from a S3 bucket is supported
- this can be enhanced to e.g. ipinfo parts

With S3 as alternative to local storage, Kubernetes-Like deployment (see e.g. #891) are made easier, as the components itself don't need persistent storage.

Still open:
- Documentation
- possibly more tests


vincentbernat:
Note I didn't have a look as I understand this is still work in progress. Tell me when you need some comments.

margau:
Hi,
feature-side, I'd consider this ready and would like to get a review.
The only thing missing is better test coverage, and the documentation of the final config structure.

vincentbernat:
Sorry for taking some time. I am fine with the feature, however I have a few remarks on the implementation.

1. At first, I thought the way the configuration is split in two parts was strange. But I understand you want to reuse configuration for several dicts (or later more stuff), so ack for this.
2. There is a mechanism to handle polymorphism, take a look at use of `helpers.ParametrizedConfiguration*` invocations. You can then have `file`, `http`, `s3`, and `s3mock` backends (to replace the `Mock` attribute, the `s3mock` can be tagged `!release` to not be exposed to the user).

indy-independence:
If using this, how/when is new updated data read from the sources?

tobikris:
I am taking over this PR, thanks for your remarks.
I first did a rebase and now the test `TestCustomDictHTTPEndpoints` is failing. I tried adding a clickhouse component to the dependencies to fix the nil pointer, but then the test case just hangs. I hope that you have a better idea of the testing helpers and what we need to initialize?

What do you think of adding the S3 dicts to the docker-compose setup? I feel like this helps understanding for users, but am not sure how to incorporate it in the setup. Is it ok like this?

I will take a look at the polymorphism remark next.

tobikris:
> If using this, how/when is new updated data read from the sources?

It is implemented to fetch the source for each request.

tobikris:
I tried to switch to the polymorphism mechanism, but am struggling with finding the correct location to put the logic in. My current state is at https://github.com/exaring/akvorado/commit/2aabbd1dccbab6ab2fa153bc3eb05072ed0839e1.

In the S3 implementation I need access to the root component and its S3 component dependency, but that seems awkward. Do you have a better approach in mind?
Feel free to either give me a hint or add your own idea to this PR here.

Thanks!

vincentbernat:
Hello Tobias, sorry for the late answer. I'll have to postpone a bit more my answers as I have currently little time to handle that.

PR #1002, opened 2023-12-18
orchestrator/kafka: alter topic configuration only when needed (config diff)

Itaah:
This PR allows orchestrator to alter Kafka topic configuration only if config has changed, instead of every time except a creation.

This is mandatory for Kafka brokers that does not allow specific config entries alter, and is a good optimization when provider sets some config entries not managed by Akvorado, to avoid unnecessary alters.

### DIFF
--- a/orchestrator/kafka/root.go
+++ b/orchestrator/kafka/root.go
@@ -105,12 +105,14 @@ func (c *Component) Start() error {
 			l.Warn().Msgf("mismatch for replication factor: got %d, want %d",
 				topic.ReplicationFactor, c.config.TopicConfiguration.ReplicationFactor)
 		}
-		if err := admin.AlterConfig(sarama.TopicResource, c.kafkaTopic, c.config.TopicConfiguration.ConfigEntries, false); err != nil {
-			l.Err(err).Msg("unable to set topic configuration")
-			return fmt.Errorf("unable to set topic configuration for %q: %w",
-				c.kafkaTopic, err)
+		if ShouldAlterConfiguration(c.config.TopicConfiguration.ConfigEntries, topic.ConfigEntries, c.config.TopicConfiguration.ConfigEntriesStrictSync) {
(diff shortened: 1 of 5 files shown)


orchestrator/kafka/root.go
```diff
@@ -105,12 +106,15 @@ func (c *Component) Start() error {
...
-			return fmt.Errorf("unable to set topic configuration for %q: %w",
-				c.kafkaTopic, err)
+		if !helpers.IsMapSubset(topic.ConfigEntries, c.config.TopicConfiguration.ConfigEntries) ||
+			(c.config.TopicConfiguration.ConfigEntriesStrictSync && helpers.Diff(topic.ConfigEntries, c.config.TopicConfiguration.ConfigEntries) != "") {
+			if err := admin.AlterConfig(sarama.TopicResource, c.kafkaTopic, c.config.TopicConfiguration.ConfigEntries, false); err != nil {
+				l.Err(err).Msg("unable to set topic configuration")
+				return fmt.Errorf("unable to set topic configuration for %q: %w",
+					c.kafkaTopic, err)
+			}
+			l.Info().Msg("topic updated")
```
vincentbernat, line 116: Add a function `shouldAlterConfiguration(target, source map[string]*string, strict bool) bool` and use it as a condition. Then, add unit tests for this function. Also, I would prefer to not use `helpers.Diff()`, nor an helper using reflection (they are hard to completely test), just write some ad-hoc code:
```go
for k, v := range target {
  if ov, ok := source[k]; !ok || *ov != *v {
    return true
  }
}
if !strict {
  return false
}
return len(target) != len(source)
```

vincentbernat (changes requested):

orchestrator/kafka/root.go
Itaah: You're right, I'll keep it simple as suggested, thanks :)

vincentbernat:
Thanks!

PR #1037, opened 2024-01-11
inlet/metadata/static: allow exporters configuration refresh from http remote data source

Itaah:
This PR aims to allows static exporters metadata refresh from remote HTTP source in addition of config based metadata.

The feature is greatly inspired from network-sources in orchestrator/clickhouse, so the PR is organized in 2 steps:

* Refactor all generic features from orchestrator/clickhouse/source (for network-sources) in a common component named remotedatasourcefetcher.
* Allow inlet/metadata/static to configure exporter-sources from HTTP endpoints, using the common/remotedatasourcefetcher component.

### DIFF
--- a/inlet/metadata/provider/static/source.go
+++ b/inlet/metadata/provider/static/source.go
@@ -1,0 +1,105 @@
+package static
+
+import (
+	"context"
+
+	"akvorado/common/helpers"
+	"akvorado/common/remotedatasourcefetcher"
+	"akvorado/inlet/metadata/provider"
+)
+
+type exporterInfo struct {
+	ExporterSubnet string
+
+	Name string `validate:"required"`
+	// Default is used if not empty for any unknown ifindexes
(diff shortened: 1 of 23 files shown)


inlet/metadata/provider/static/source.go
```diff
@@ -0,0 +1,107 @@
...
+		if id == "static" {
+			continue
+		}
+		for _, exporterData := range results {
+			exporterSubnet, err := helpers.SubnetMapParseKey(exporterData.ExporterSubnet)
+			if err != nil {
+				p.r.Err(err).Msg("failed to decode subnet")
+				continue
+			}
+			// Concurrency for same Exporter config across multiple remote data sources is not handled
```
vincentbernat, line 87: Keep the write lock? This does not seem to be a hot path.

vincentbernat:
This looks good to me. There are two small modifications and we can merge.

PR #1039, opened 2024-01-17
Add TLS support for ClickhouseDB

netixx:
Add support for TLS connection for clickhouse.

Code merged between kafka and clickhouse since it's exactly the same *tls.Config object that is required in both cases.


common/kafka/config.go
```diff
@@ -28,22 +25,12 @@ type Configuration struct {
...
-	// CAFile tells the location of the CA certificate to check broker
-	// certificate. If empty, the system CA certificates are used instead.
-	CAFile string // no validation as the orchestrator may not have the file
-	// CertFile tells the location of the user certificate if any.
-	CertFile string `validate:"required_with=KeyFile"`
-	// KeyFile tells the location of the user key if any.
-	KeyFile string
+// TLSAndSASLConfiguration defines TLS configuration.
+type TLSAndSASLConfiguration struct {
+	helpers.TLSConfiguration `mapstructure:",squash"`
```
vincentbernat, line 33: You need also `yaml:",inline"` (as the orchestrator may republish it as YAML).

vincentbernat (changes requested):

common/kafka/config.go
netixx: oh, missed that :) 

done!

PR #1041, opened 2024-01-17
Support providing exporter and iface metadata through metadata instead of classifiers

netixx:
Sometime exporter name and interface description do not carry all the required information for classification and metadata extraction, supporting a way to provide the data through metadata compoenent (only static seems to make sense at this points) enables more use-cases.

### DIFF
--- a/inlet/core/enricher.go
+++ b/inlet/core/enricher.go
@@ -36,19 +36,19 @@ func (c *Component) enrichFlow(exporterIP netip.Addr, exporterStr string, flow *
 			c.metrics.flowsErrors.WithLabelValues(exporterStr, "SNMP cache miss").Inc()
 			skip = true
 		} else {
-			flowExporterName = answer.ExporterName
-			expClassification.Region = answer.ExporterRegion
-			expClassification.Role = answer.ExporterRole
-			expClassification.Tenant = answer.ExporterTenant
-			expClassification.Site = answer.ExporterSite
-			expClassification.Group = answer.ExporterGroup
+			flowExporterName = answer.Exporter.Name
+			expClassification.Region = answer.Exporter.Region
+			expClassification.Role = answer.Exporter.Role
+			expClassification.Tenant = answer.Exporter.Tenant
+			expClassification.Site = answer.Exporter.Site
+			expClassification.Group = answer.Exporter.Group
 			flowInIfIndex = flow.InIf
-			flowInIfName = answer.InterfaceName
-			flowInIfDescription = answer.InterfaceDescription
-			flowInIfSpeed = uint32(answer.InterfaceSpeed)
-			inIfClassification.Provider = answer.InterfaceProvider
-			inIfClassification.Connectivity = answer.InterfaceConnectivity
-			inIfClassification.Boundary = answer.InterfaceBoundary
+			flowInIfName = answer.Interface.Name
+			flowInIfDescription = answer.Interface.Description
+			flowInIfSpeed = uint32(answer.Interface.Speed)
+			inIfClassification.Provider = answer.Interface.Provider
+			inIfClassification.Connectivity = answer.Interface.Connectivity
+			inIfClassification.Boundary = answer.Interface.Boundary
 			flowInIfVlan = flow.SrcVlan
 		}
 	}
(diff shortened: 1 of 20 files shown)


inlet/metadata/provider/root.go
```diff
@@ -32,8 +36,18 @@ type BatchQuery struct {
...
+	ExporterRole          string
+	ExporterTenant        string
+	ExporterSite          string
+	ExporterGroup         string
+	InterfaceName         string
+	InterfaceDescription  string
+	InterfaceSpeed        uint
+	InterfaceProvider     string
+	InterfaceConnectivity string
+	InterfaceBoundary     schema.InterfaceBoundary
```
vincentbernat, line 50: Having substructures was cleaner, wasn't it? Notably if you didn't do that because of the `string`/`InterfaceBoundary` mismatch.

inlet/metadata/provider/root.go
```diff
@@ -9,13 +9,17 @@ import (
...
 type Interface struct {
-	Name        string `validate:"required"`
-	Description string `validate:"required"`
-	Speed       uint   `validate:"required"`
+	Name         string `validate:"required"`
+	Description  string `validate:"required"`
+	Speed        uint   `validate:"required"`
+	Provider     string
+	Connectivity string
+	Boundary     string
```
vincentbernat, line 22: string? This could be a `schema.InterfaceBoundary`. I suppose this is because it needs to be used in configuration, but that's something you can handle by registering a deserializer.

inlet/core/enricher.go
```diff
@@ -207,7 +226,11 @@ func (c *Component) writeExporter(flow *schema.FlowMessage, classification expor
 	return true
 }
 
-func (c *Component) classifyExporter(t time.Time, ip string, name string, flow *schema.FlowMessage) bool {
+func (c *Component) classifyExporter(t time.Time, ip string, name string, flow *schema.FlowMessage, classification exporterClassification) bool {
+	// we already have the info provided by the metadata component
+	if classification.Group != "" || classification.Role != "" || classification.Site != "" || classification.Region != "" || classification.Tenant != "" {
```
vincentbernat, line 231: This could be `classification != exporterClassification{}` (I think)

inlet/metadata/provider/snmp/poller_test.go
```diff
@@ -226,8 +226,8 @@ func TestPoller(t *testing.T) {
 			if diff := helpers.Diff(got, []string{
 				fmt.Sprintf(`%s exporter62 641 Gi0/0/0/0 Transit 10000`, exporterStr),
 				fmt.Sprintf(`%s exporter62 642 Gi0/0/0/1 Peering 20000`, exporterStr),
-				fmt.Sprintf(`%s exporter62 643   0`, exporterStr), // negative cache
-				fmt.Sprintf(`%s exporter62 644   0`, exporterStr), // negative cache
+				fmt.Sprintf(`%s exporter62 643 Gi0/0/0/2  10000`, exporterStr), // no ifAlias
```
vincentbernat, line 229: I don't understand why this changed?

vincentbernat (changes requested):

inlet/metadata/provider/root.go
netixx: Actually I wasn't sure exactly how to do that properly, especially since it's required for all kinds of sources (ie. config (mapstructure decode ?), http sources (custom code after JQ call ?) and when config is given to inlet from orchestrator (yaml ?).

Also, I didn't like that the `iota` value would be sent over the network from orchestrator to inlet (maybe there is a version missmatch, and the two values don't align) ?

How do you think this could be done properly ?

inlet/metadata/provider/root.go
netixx: That was the main reason yes :) 

Also I figured this was one less struct copied around (dubious argument I agree !).

In the end, I also think that it's complicated to share structs between user input and internal code needs.
For example, there is `ExporterConfiguration` for internal needs (with ifindexes lookup),
but it cannot be loaded from json because json does not support `map[uint]`, only `map[string]`, so there is `exporterInfo` with an `[]exporterInterfaces` (which is basically the same as Interface, but with the ifindex pushed down). In the same way, `exporterInfo` is also required because jq requires that we end up with a list of exporters, and not a map (`ExporterSubnet` is pushed down compared to loading from the config).
But at the same time, this leads to somewhat duplicated code...

inlet/core/enricher.go
netixx: I guess (not sure) this is an extra memory allocation on the hot path ?

vincentbernat: Being a const, I suppose the compiler would do a good job doing the same thing as the expanded version. It's just the day you add something to this struct, it's one less site to modify.

inlet/metadata/provider/snmp/poller_test.go
vincentbernat: Ack for this one.

inlet/metadata/provider/root.go
vincentbernat: I agree with your last paragraph: let's not try that if it introduces more complexity. Let me think about the other comment.

inlet/metadata/provider/root.go
vincentbernat: For me, everything should be handled by something like this: https://github.com/akvorado/akvorado/blob/main/common/kafka/config.go#L80-L124.

inlet/metadata/provider/root.go
vincentbernat: Note that you can definitely load anything from JSON by implementing `UnmarshalJSON()` method. It really depends on what solution is the simplest. We can postpone this question to later too, as long as the test coverage is OK, I should be able to change it if it bothers me too much.

inlet/core/enricher.go
netixx: done

inlet/metadata/provider/root.go
netixx: done as well, using your example as a baseline (I never used UnmarshalText before :))

netixx:
I don't know why the tests are broken all of a sudden (seems ok on my laptop).
I keep getting this error though : 
```
common/clickhousedb/tests.go:47:28: cannot use ctrl (variable of type *"go.uber.org/mock/gomock".Controller) as *"github.com/golang/mock/gomock".Controller value in argument to mocks.NewMockConn
```
but it was already the case before so I guess/hope it's not related to my changes ?

vincentbernat:
This is related to https://github.com/akvorado/akvorado/commit/07922903161137557da70f05c6fe80e21e4d4656. You should delete `bin/mockgen`.

vincentbernat:
The Kafka error is likely because, again, Bitnami changes something important in the latest version of a stable image... Let me merge this as is and check myself.

PR #1043, opened 2024-01-20
inlet/metadata: gNMI support

vincentbernat:
This was a bit difficult, but thanks to @hellt and @karimra for their help!

It was only tested with SR Linux, but with a bit of luck, it should work for other equipments.


hellt:
@vincentbernat can you please share the dummy configuration for the gnmi-based inlet config?
We seem to have some issues with getting the `targets` to work, which is a map, keyed by the target.

```
Error: invalid configuration:
invalid key "Inlet[0].Metadata.Providers[0].Config.Targets.10.0.0.3"
```

vincentbernat:
First, `models` is to add new models or redefine existing ones, it's not to select the model you want. The first matching model is automatically selected.

I have fixed the bug in 296178459b31. The Docker image `:main` should be ready soon. Thanks for testing! :-)

Issue #62, opened 2022-07-29
Add `SrcCity` and `DstCity`


vincentbernat:
In addition to `SrcCountry` and `DstCountry`, it would be nice to be able to add `SrcCity` and `DstCity`.

However, this may a rabbit hole. The City Maxmind databases also contains:
- city
- continent
- country
- location (longitude/latitude)
- postal code
- subdivisions (many)

For each data, we may do something different:
- continent: get it from the country code using a dictionary lookup at query time
- city: use geoid
- country: use geoid or keep doing what we do
- location: consider city geoid is enough
- postal code: consider city geoid is enough
- subdivisions: use geoid

We could replace `SrcCountry` and `DstCountry` by `SrcGeoID` and `DstGeoID` and have a dictionary lookup during query to extract all the other fields we want (`SrcGeoContinent`, `SrcGeoCountry`, `SrcGeoRegion1`, `SrcGeoRegion2`, `SrcGeoCity`) using aliases (that's what is done for `PacketSize`, it's free to add more).

It would mean that we need access to GeoIP database from inlet (like currently) and from ClickHouse. To simplify things, we could consider moving all the geolocation stuff into ClickHouse. During ingestion, it would lookup the geoname ID. And during query, you get the remaining stuff.

AS numbers would stay in Akvorado as there are various ways we can get them and in the future, we want to configure their sources accurately.

- [ ] Move `inlet/geoip` to `common/geoip`, adapt `inlet` to use it (drop `LookupCountry`)
- [ ] Make `common/geoip` able to dump the provided database to CSV
- [ ] Make `orchestrator/clickhouse` expose this CSV to ClickHouse
- [ ] Add dictionary to ClickHouse, materialize a `SrcGeoID`/`DstGeoID` columns, drop the `SrcCountry`/`DstCountry`, add aliases for `SrcGeo*`/`DstGeo*` columns
- [ ] Make the new columns/aliases available from the console

It should be noted that using the city database will have an impact on the efficiency of the aggregation (it should be mentioned in documentation).

netixx:
We had an idea similar to this in scope, but different in implementation.

The idea was to have everything done through a lookup in the networks dictionary inside clickhouse only.
We have already thought quite a lot about this, so here is how it would go in details:
- load one or more geoip database inside the orchestrator
- iter the database, and inject their contents into a trie
- iter the contents of the existing "networks", and put them into the same trie
- iter this converged trie, and fill the attributes that are undefined (eg. city) from the parent
[...]

PR #1059, opened 2024-01-25
Do geoip enrich in clickhouse instead of inlet

netixx:
One solution to https://github.com/akvorado/akvorado/issues/62.

Here we basically implement what is defined in this issue.

Opening this as draft first as I still have a few points to iron out, but I would like some feedback first:
- Should we decommission the inlet way of enriching the data ? This would simplify the way we handle the schema (conditional columns), but there would be a breaking change in terms of configuration (geoip config moving to orchestrator component, supporting multiple databases). Note that there is an actual difference: this way of doing enrich puts extra load on clickhouse (but less on inlet), we tested clickhouse with 2M networks, and didn't have any issue so I think it's ok.
- The geoip component runs on it's own, and is passed as a dependency of orchestrator. There needs to be a way for geoip to notify orchestrator when an update is available, this is done through ad-hoc channels, but maybe we should have something more structured - especially if we want to extend this mechanism to other components (something like an event but or pub/sub lib to handle communication between components). Or do something else entirely :).
- Meaning of Site/Region/Country: it looks to me like this was unclear in the examples before (esp. Region). With this change we can now define Src/DstNetCountry from the custom networks, which I think is good (for internal networks for example, the country in geoip could be less accurate).
- We refresh the whole "convergedNetworks" trie each time there is any change, which is wastefull but at the same time simpler than to try to prune dead branches from the trie... We plan to implement a way for the source component (http fetcher, geoip refresh) actually trigger the change only when an actual change in the data occurs (eg. with a hash), which I think is good enough (orchestrator is not in the hot path anyways).

I will push a doc update once we have converged on the points that remain undefined (and other feedback).


vincentbernat:
1. Yes, let's decommission the component in the inlet. I am not worried about breaking anything for most users, because it is easy to add a few lines of code to migrate the configuration from inlet to orchestrator (this would be the first time moving from one component to another, but there is code existing doing that inside a component, this should only be done somewhere in `cmd`). We can focus on that later. However, this would also mean we have to handle AS numbers.
2. My philosophy is that on the first occurrence, use something adhoc. Wait for the second occurrence to refactor. I don't want extra complexity for something that will never be used. Also, in this case, you can just make one component depend on the other and use a function call to notify the orchestrator.
3. Site/Region were words to not collide with things like City/Country on purpose. For us, regions are France, Italy, Caraibes, Reunion, Mayotte, so there is a mix of countries, departments and vague regions. So, I would propose to have `DstGeoCity`, `DstGeoCountry` and `DstNetSite`, `DstNetRegion`, this does not prevent to override `DstGeoCity` with something internal like you propose.
4. This can be postponed I think.

netixx:
Noted :) 


While looking at removing geoip from inlet, I saw
```
const (
	// ASNProviderFlow uses the AS number embedded in flows.
	ASNProviderFlow ASNProvider = iota
	// ASNProviderFlowExceptPrivate uses the AS number embedded in flows, except if this is a private AS.
	ASNProviderFlowExceptPrivate
	// ASNProviderGeoIP pulls the AS number from a GeoIP database.
	ASNProviderGeoIP
	// ASNProviderRouting uses the AS number from BMP
	ASNProviderRouting
	// ASNProviderRoutingExceptPrivate uses the AS number from BMP, except if this is a private AS.
	ASNProviderRoutingExceptPrivate
)
```

Which means that we need to do something with the `ASNProviderGeoIP` option (I guess remove it).
Although it's not accurate to rely on source routing lookups to get AS-PATH info, I think it should be fine for the origin AS (except in case of hijack of course :) ).

netixx:
Regarding 3., so the goal is to have "custom" location components (that are user-defined in the end) : Site and Region
And have "standard" (defined by geoip) location components: City, Country (and I would like to add State as well)

Should we rename SrcCountry to SrcGeoCountry ? Or use SrcCity to be consistent ?
For state, I am stuck on the naming, because state doesn't make sense in every country. Maxmind uses "subdivisions" which works for everything, but is not really user-friendly.

vincentbernat:
I would favor renaming (but we can do that in a subsequent commit before release). Using state even if Maxmind says subdivisions should be fine too. If people are confused, we can still add a note in the documentation

vincentbernat:
For AS, yes, GeoIP should be removed. We won't be able to express exactly the same configuration when this is done in ClickHouse instead, but I think this should still be OK (I think GeoIP only makes sense as the last parameter anyway). As a first step, in ClickHouse, using GeoIP is done when current AS is 0.

netixx:
> For AS, yes, GeoIP should be removed. We won't be able to express exactly the same configuration when this is done in ClickHouse instead, but I think this should still be OK (I think GeoIP only makes sense as the last parameter anyway). As a first step, in ClickHouse, using GeoIP is done when current AS is 0.

Ok, I wasn't sure it was possible for a column to reference itself in the expression, like
`if(SrcAs == 0, dictGetOrDefault('networks', 'asn', SrcAddr, 0), SrcAs)`
I guess this is applied between the kafka table and the final tables ?

netixx:
ok, so I used
```
{
Key:                    ColumnSrcAS,
ClickHouseType:         "UInt32",
ClickHouseGenerateFrom: "if(SrcAS == 0, dictGetOrDefault('networks', 'asn', SrcAddr, 0), SrcAS)",
},
```
as column definition, but this breaks encoding of the AS in protobuf because of
```
if column.ClickHouseTransformFrom != nil ||
	column.ClickHouseGenerateFrom != "" ||
	column.ClickHouseAlias != "" {
	column.ProtobufIndex = -1
	continue
}
```
that generates a protobuf ID for the field.

I added a check to detect that the column references itself inside the "ClickHouseGenerateFrom".

I will test this version further to see what else is going to go wrong :)

netixx:
Ok I got an error creating the table:
```
akvorado-orchestrator-1  | {"level":"error","error":"cannot create raw flows consumer view: code: 47, message: Missing columns: 'SrcAS' 'DstAS' while processing query: 'WITH arrayCompact(DstASPath) AS c_DstASPath SELECT TimeReceived, SamplingRate, ExporterAddress, ExporterName, ExporterGroup, ExporterRole, ExporterSite, ExporterRegion, ExporterTenant, SrcAddr, DstAddr, SrcNetMask, DstNetMask, if(SrcAS = 0, dictGetOrDefault('akvorado.networks', 'asn', SrcAddr, 0), SrcAS) AS SrcAS, if(DstAS = 0, dictGetOrDefault('akvorado.networks', 'asn', DstAddr, 0), DstAS) AS DstAS, dictGetOrDefault('akvorado.networks', 'name', SrcAddr, '') AS SrcNetName, dictGetOrDefault('akvorado.networks', 'name', DstAddr, '') AS DstNetName, dictGetOrDefault('akvorado.networks', 'role', SrcAddr, '') AS SrcNetRole, dictGetOrDefault('akvorado.networks', 'role', DstAddr, '') AS DstNetRole, dictGetOrDefault('akvorado.networks', 'site', SrcAddr, '') AS SrcNetSite, dictGetOrDefault('akvorado.networks', 'site', DstAddr, '') AS DstNetSite, dictGetOrDefault('akvorado.networks', 'region', SrcAddr, '') AS SrcNetRegion, dictGetOrDefault('akvorado.networks', 'region', DstAddr, '') AS DstNetRegion, dictGetOrDefault('akvorado.networks', 'tenant', SrcAddr, '') AS SrcNetTenant, dictGetOrDefault('akvorado.networks', 'tenant', DstAddr, '') AS DstNetTenant, dictGetOrDefault('akvorado.networks', 'country', SrcAddr, '') AS SrcCountry, dictGetOrDefault('akvorado.networks', 'country', DstAddr, '') AS DstCountry, dictGetOrDefault('akvorado.networks', 'city', SrcAddr, '') AS SrcGeoCity, dictGetOrDefault('akvorado.networks', 'city', DstAddr, '') AS DstGeoCity, dictGetOrDefault('akvorado.networks', 'state', SrcAddr, '') AS SrcGeoState, dictGetOrDefault('akvorado.networks', 'state', DstAddr, '') AS DstGeoState, DstASPath, c_DstASPath[1] AS Dst1stAS, c_DstASPath[2] AS Dst2ndAS, c_DstASPath[3] AS Dst3rdAS, DstCommunities, arrayMap((asn, l1, l2) -> ((bitShiftLeft(CAST(asn, 'UInt128'), 64) + bitShiftLeft(CAST(l1, 'UInt128'), 32)) + CAST(l2, 'UInt128')), DstLargeCommunitiesASN, DstLargeCommunitiesLocalData1, DstLargeCommunitiesLocalData2) AS DstLargeCommunities, InIfName, OutIfName, InIfDescription, OutIfDescription, InIfSpeed, OutIfSpeed, InIfConnectivity, OutIfConnectivity, InIfProvider, OutIfProvider, InIfBoundary, OutIfBoundary, EType, Proto, SrcPort, DstPort, Bytes, Packets, ForwardingStatus FROM akvorado.flows_LAABIGYMRYZPTGOYIIFZNYDEQM_raw WHERE length(_error) = 0', required columns: 'SamplingRate' 'SrcAddr' 'DstASPath' 'ExporterTenant' 'ExporterRole' 'ForwardingStatus' 'OutIfSpeed' 'InIfName' 'TimeReceived' 'ExporterRegion' 'ExporterSite' 'DstPort' 'DstCommunities' 'DstAddr' 'OutIfName' 'InIfConnectivity' 'SrcNetMask' 'ExporterName' 'DstLargeCommunitiesASN' 'InIfDescription' 'OutIfDescription' 'ExporterGroup' 'ExporterAddress' 'OutIfConnectivity' 'OutIfProvider' 'Proto' 'OutIfBoundary' 'InIfBoundary' 'EType' 'DstAS' 'Packets' 'SrcPort' 'DstNetMask' 'InIfSpeed' 'Bytes' 'InIfProvider' 'SrcAS' 'DstLargeCommunitiesLocalData1' '_error' 'DstLargeCommunitiesLocalData2', maybe you meant: 'SamplingRate', 'SrcAddr', 'DstASPath', 'ExporterTenant', 'ExporterRole', 'ForwardingStatus', 'OutIfSpeed', 'InIfName', 'TimeReceived', 'ExporterRegion', 'ExporterSite', 'DstPort', 'DstCommunities', 'DstAddr', 'OutIfName', 'InIfConnectivity', 'SrcNetMask', 'ExporterName', 'DstLargeCommunitiesASN', 'InIfDescription', 'OutIfDescription', 'ExporterGroup', 'ExporterAddress', 'OutIfConnectivity', 'OutIfProvider', 'Proto', 'OutIfBoundary', 'InIfBoundary', 'EType', 'Packets', 'SrcPort', 'DstNetMask', 'InIfSpeed', 'Bytes', 'InIfProvider', 'DstLargeCommunitiesLocalData1' or 'DstLargeCommunitiesLocalData2'","time":"2024-01-30T09:52:53Z","caller":"akvorado/orchestrator/clickhouse/root.go:111","module":"akvorado/orchestrator/clickhouse","message":"database migration error"} 
```

it looks like it's not supported for a column to reference itself (circular) https://github.com/ClickHouse/ClickHouse/pull/25059.

So I guess we need a temporary column that exists only in the kafka table ?

vincentbernat:
I really think this is possible. Maybe try to name the FROM:
```
CREATE MATERIALIZED VIEW ...
AS SELECT
...
if(o.SrcAS == 0, dictGetOrDefault('networks', 'asn', o.SrcAddr, 0), o.SrcAS) AS SrcAS
...
FROM default.flows_ZUYGDTE3EBIXX352XPM3YEEFV4_raw o
```

netixx:
I found what the issue was: the raw flows table didn't actually have the Src/DstAS columns, since they where generatedFrom, so I added a check to see if they were self-generated.

netixx:
Ok, I think we can move forward with this PR.

I still have the following issues:
- race test on github seems to fail, but I can't reproduce with `make test-race` locally (no FAIL there)
- need a missing bit of code to migrate configuration from inlet to orchestrator. Do you have an example in mind I could adapt ?
- I found an issue with the reload dictionary: it takes some time for traefik to register the route, and so when we sync the networks initially, clickhouse gets a 404. I think I should add a exponential backoff thingy, but this will also delay the startup
- I loaded the dict, with IPINFO database into clickhouse, and that's around 4M entries.
- there is a schema change with this PR, due to the removal of SrcCountry and DstCountry. Still running some tests, but at first glance, it seems I couldn't get the clickhouse to consume the new format from kafka (not really sure what the issue is, since it ran OK in docker-compose).

vincentbernat:
Don't bother about the migration yet. `inlet/metadata/provider/snmp/config.go`, `ConfigurationUnmarshallerHook()`.

netixx:
I could finally test this further.

We'll need to wait for https://github.com/kentik/patricia/pull/31 to be merged before proceeding (I need a way to get all the tags from the root of the tree to the leaf, so that I can push them down to the prefix in networks.csv)

vincentbernat:
They can be a bit slow. You can just point to your pull request.

```
replace github.com/kentik/patricia => github.com/netixx/patricia v0.0.0-20220731174804-18c23f58bc16
```

To get the right reference.

```
go get github.com/netixx/patricia@yourcommit
```

Makefile
```diff
@@ -206,3 +206,7 @@ help:
 .PHONY: version
 version:
 	@echo $(VERSION)
+
+.PHONY: build-docker
+build-docker:
```
vincentbernat, line 211: `make docker`

common/schema/definition.go
```diff
@@ -228,69 +243,113 @@ func flows() Schema {
...
+				ParserType:             "string",
+				ClickHouseType:         "LowCardinality(String)",
+				ClickHouseGenerateFrom: fmt.Sprintf("dictGetOrDefault('%s', 'state', SrcAddr, '')", DictionaryNetworks),
+			},
+			{
+				Key:                    ColumnDstGeoState,
+				ParserType:             "string",
+				ClickHouseType:         "LowCardinality(String)",
+				ClickHouseGenerateFrom: fmt.Sprintf("dictGetOrDefault('%s', 'state', DstAddr, '')", DictionaryNetworks),
+			},
```
vincentbernat, line 352: State and City should be optional. Otherwise, people will pay the lookup cost even when there is no database loaded for that.

common/schema/definition.go
```diff
@@ -509,6 +568,16 @@ END`,
 	}.finalize()
 }
 
+func (column *Column) isSelfGenerated() bool {
+	return strings.Contains(column.ClickHouseGenerateFrom, column.Name)
+}
```
vincentbernat, line 573: It should be a specific attribute `ClickHouseSelfGenerated` that should be set on the few columns in this case. While unlikely to have a name of an unrelated column, this is safer.

docker/docker-compose.yml
```diff
@@ -64,14 +64,18 @@ services:
       test: ["CMD-SHELL", "timeout 3 redis-cli ping | grep -q PONG"]
 
   akvorado-service: &akvorado-image
-    image: ghcr.io/akvorado/akvorado:latest
+    image: akvorado:latest
```
vincentbernat, line 67: To be reverted (and maybe update the Makefile to use the right name).

docker/docker-compose.yml
```diff
@@ -130,7 +130,7 @@ services:
     volumes:
       - ../orchestrator/clickhouse/data/docker-entrypoint.sh:/entrypoint-override.sh:ro
       - akvorado-clickhouse:/var/lib/clickhouse
-    entrypoint: /entrypoint-override.sh
+    # entrypoint: /entrypoint-override.sh
```
vincentbernat, line 133: Should still be there?

common/helpers/subnetmap.go
```diff
@@ -13,6 +13,7 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/kentik/patricia"
+	"github.com/kentik/patricia/generics_tree"
 	tree "github.com/kentik/patricia/generics_tree"
```
vincentbernat, line 17: Duplicate. Replace occurrences of `generics_tree` by `tree`.

common/helpers/subnetmap.go
```diff
@@ -57,25 +58,69 @@ func (sm *SubnetMap[V]) ToMap() map[string]V {
...
+	if bits != 128 {
+		return fmt.Errorf("%q is not an IPv6 subnet", ipNet)
+	}
+	plen, _ := ipNet.Mask.Size()
+	sm.tree.Set(patricia.NewIPv6Address(ipNet.IP.To16(), uint(plen)), v)
+	return nil
+}
+
+// Update inserts the given key k into the SubnetMap, calling updateFunc with the existing value.
+func (sm *SubnetMap[V]) Update(k string, v V, updateFunc generics_tree.UpdatesFunc[V]) error {
```
vincentbernat, line 82: `generics_tree` → `tree`

orchestrator/clickhouse/geoip/database.go
```diff
@@ -36,13 +53,23 @@ func (c *Component) openDatabase(which string, path string, container *atomic.Po
...
+		c.db.geo[path] = newOne
+	}
+	c.db.lock.Unlock()
 	c.metrics.databaseRefresh.WithLabelValues(which).Inc()
 	if oldOne != nil {
 		c.r.Debug().
 			Str("database", path).
 			Msgf("closing previous %s database", which)
-		(*oldOne).Close()
+		oldOne.Close()
```
vincentbernat, line 72: I think there is a race condition. You need to keep the write lock up to closing...

orchestrator/clickhouse/geoip/iter.go
```diff
@@ -0,0 +1,46 @@
...
+// IterASNDatabase iter all entries in the given asn database path.
+func (c *Component) IterASNDatabase(path string, f AsnIterFunc) error {
+	c.db.lock.RLock()
+	geoDB := c.db.asn[path]
+	c.db.lock.RUnlock()
+	if geoDB != nil {
+		return geoDB.IterASNDatabase(f)
+	}
+
+	return fmt.Errorf("database not found %s", path)
```
vincentbernat, line 45: And the read lock until the end. Otherwise, the database can be closed while being used.

orchestrator/clickhouse/geoip/root.go
```diff
@@ -0,0 +1,267 @@
+// SPDX-FileCopyrightText: 2022 Free Mobile
```
vincentbernat, line 1: Note that you can put yourself/own company here.

orchestrator/clickhouse/root.go
```diff
@@ -74,13 +88,21 @@ func New(r *reporter.Reporter, configuration Configuration, dependencies Depende
...
 
 // Start the ClickHouse component.
 func (c *Component) Start() error {
 	c.r.Info().Msg("starting ClickHouse component")
 
+	// stub to prevent tomb dying immediately after migrations are done
+	c.t.Go(func() error {
+		<-c.t.Dying()
+		return nil
+	})
```
vincentbernat, line 104: Yes, that's expected. Nothing much to do, this is part of the design of the tombs.

vincentbernat (changes requested):
Looks almost good. On my side, I will still need to do the migration.

vincentbernat:
Also, later, we may introduce an optimization to only have one lookup per key.

Makefile
netixx: renamed

common/schema/definition.go
netixx: done

docker/docker-compose.yml
netixx: removed (leftover from my testing)

orchestrator/clickhouse/geoip/iter.go
netixx: done

common/schema/definition.go
netixx: do you mean optional as they can activate the column on demand via config ?

Should it be linked to whether or not the geoip enrich is configured or not ?

common/helpers/subnetmap.go
netixx: done (I let auto-import catch me :) )

orchestrator/clickhouse/root.go
vincentbernat: I think they have an explanation here: https://github.com/go-tomb/tomb/issues/16 (but right now, I don't really grasp the why).

vincentbernat:
Tell me if you want to do the optimization while I work on the configuration migration. Thanks for rebasing BTW, I forgot to check if I should squash or rebase.

vincentbernat:
I have pushed the migration and the documentation.

vincentbernat:
Did you ingest a geoip database containing cities? I would be interested in the memory usage for the dictionary inside ClickHouse: https://clickhouse.com/docs/en/operations/system-tables/dictionaries#system_tables-dictionaries.

vincentbernat:
Another question. Can you explain if the reload of the networks dictionary should really be done synchronously and why we should wait for first sync for the component to be considered to be started? This adds a substantial time to tests since the HTTP address is bogus when testing migrations.

vincentbernat:
I have pushed 93452cfbfa4c that could affect you.

Using AS names for tenants broke our setup. Not using them may break yours. Maybe this could be configurable instead.

This also raises the question about overriding more specific entries. When you have in a source a prefix A/16 with country=B and a geoip entry for A/16 with country=C, you get A/16 and country=B. However, if a more specific entries appear, like A/17 with country=C, you get A/17 and country=C, while you might want A/17 and country=B. Maybe each source should be configurable to override or not the previous ones.

PR #1060, opened 2024-01-26
Add config hint for the inlet service

netfreak98:
Added a hint to mitigate AS0 related issues


vincentbernat:
Instead, can you put a paragraph in the troubleshooting documentation (or I can do it)? I don't want to litter the documentation files with hints on how to adapt the configuration to a specific configuration.

vincentbernat:
Not needed anymore as geoip is always used as a fallback.

PR #1099, opened 2024-03-01
console/query/filter: Allow configurable LIMIT through input on complete API

Itaah:
This PR aims to add a `limit` body parameter on filter completion API, to allow third party apps loading more than default 20 values in one time.

This could also be really useful for Grafana dashboards with JSON-API based variables.


vincentbernat:
Can you rebase?

PR #1128, opened 2024-03-20
Allow using fields of the netflow packet to set the flow TimeReceived

netixx:
Today the timestamp can only be from kernel timetstamp put on the UDP packet by the kernel.

I propose to add 2 alternative methods of getting the timestamp for netflow/IPFix packets:
- TimestampSourceNetflowPacket: use the timestamp field in the netflow packet itself
- TimestampSourceNetflowFirstSwitched: use the FirstSwitched field from each flow (the field is actually in uptime, so we need to shift it according to sysUptime)

Using those fields requires the router to have accurate time (probably NTP), but it allows for architectures where a UDP packet is not immediately received by the collector, eg. if there is a kafka in-between. That in turns allows to do maintenance on the collector, without messing up the statistics

### DIFF
--- a/inlet/flow/decoder/config.go
+++ b/inlet/flow/decoder/config.go
@@ -1,0 +1,59 @@
+package decoder
+
+import (
+	"akvorado/common/helpers/bimap"
+	"errors"
+)
+
+// TimestampSource defines the method to use to extract the TimeReceived for the flows
+type TimestampSource uint
+
+const (
+	// TimestampSourceUDP tells the decoder to use the kernel time at which
+	// the UDP packet was received
+	TimestampSourceUDP TimestampSource = iota
+	// TimestampSourceNetflowPacket tells the decoder to use the timestamp
+	// from the router in the netflow packet
+	TimestampSourceNetflowPacket
+	// TimestampSourceNetflowFirstSwitched tells the decoder to use the timestamp
+	// from each flow "FIRST_SWITCHED" field
+	TimestampSourceNetflowFirstSwitched
+)
+
+var (
+	timestampSourceMap = bimap.New(map[TimestampSource]string{
+		TimestampSourceUDP: "udp",
+		TimestampSourceNetflowPacket: "netflow-packet",
+		TimestampSourceNetflowFirstSwitched:  "netflow-first-switched",
+	})
+	errUnknownTimestampSource = errors.New("unknown TimestampSource")
+)
+
+// MarshalText turns an interface boundary to text
+func (ib TimestampSource) MarshalText() ([]byte, error) {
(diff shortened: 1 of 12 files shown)


inlet/flow/decoder/netflow/decode.go
```diff
@@ -151,6 +151,8 @@ func (nd *Decoder) decodeRecord(version uint16, obsDomainID uint32, samplingRate
 		// Remaining
 		case netflow.NFV9_FIELD_FORWARDING_STATUS:
 			nd.d.Schema.ProtobufAppendVarint(bf, schema.ColumnForwardingStatus, decodeUNumber(v))
+		case netflow.NFV9_FIELD_FIRST_SWITCHED:
+			bf.TimeReceived = decodeUNumber(v) + sysOffset
```
vincentbernat, line 155: So, both additional options mean the same thing?

vincentbernat (changes requested):

inlet/flow/decoder/netflow/decode.go
netixx: not sure I get what you mean.

In the case of FIRST_SWITCHED, I use the field inside each of the flow in the flowset to set "TimeReceived", which is then used afterwards.

In the case of NetflowPacket, I use the timestamp on the netflow packet for all flows in the flowset.
The `ts` in that case is `ts = uint64(packetNFv9.UnixSeconds)`.

```
for _, fmsg := range flowMessageSet {
		if !nd.useTsFromFirstSwitched {
			fmsg.TimeReceived = ts
		}
		fmsg.ExporterAddress = exporterAddress
	}
```

In the case of the UDP timestamp, ts = timestamp from kernel (`uint64(in.TimeReceived.UTC().Unix())`).

inlet/flow/decoder/config.go
```diff
@@ -0,0 +1,59 @@
...
+	TimestampSourceNetflowFirstSwitched
+)
+
+var (
+	timestampSourceMap = bimap.New(map[TimestampSource]string{
+		TimestampSourceUDP: "udp",
+		TimestampSourceNetflowPacket: "netflow-packet",
+		TimestampSourceNetflowFirstSwitched:  "netflow-first-switched",
+	})
+	errUnknownInterfaceBoundary = errors.New("unknown TimestampSource")
```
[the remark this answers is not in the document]
netixx: fixed

inlet/flow/decoder/netflow/decode.go
vincentbernat: OK, I missed this part of the code.

vincentbernat (changes requested):
Ack for the code part. Could you add the new option in the documentation and an entry 🌱 in the changelog?

PR #1130, opened 2024-03-22
console/clickhouse: Expose API route to get best table and interval from time range

Itaah:
This PR aims to expose as an API route, the internal logic to compute the appropriate `table` and `interval` (in seconds) to use.
This is very useful for third party apps like Grafana dashboards, running Akvorado-like SQL queries, to adapt perfectly with current Akvorado configuration.

### DIFF
--- a/console/clickhouse.go
+++ b/console/clickhouse.go
@@ -13,8 +13,14 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/gin-gonic/gin"
+	"net/http"
+
+	"akvorado/common/helpers"
 	"akvorado/console/query"
 )
+
+const defaultPointsNumber = 200
 
 // flowsTable describe a consolidated or unconsolidated flows table.
 type flowsTable struct {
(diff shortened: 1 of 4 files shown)


console/clickhouse.go
```diff
@@ -140,26 +148,24 @@ func templateContext(context inputContext) string {
 	return fmt.Sprintf("context `%s`", string(encoded))
 }
 
+func (c *Component) getTableAndIntervalHandlerFunc(gc *gin.Context) {
+	input := inputContext{Points: 200}
```
vincentbernat, line 152: Move 200 to a constant so that it stays in sync.

console/root.go
```diff
@@ -99,6 +99,7 @@ func (c *Component) Start() error {
 	endpoint.GET("/filter/saved", c.filterSavedListHandlerFunc)
 	endpoint.DELETE("/filter/saved/:id", c.filterSavedDeleteHandlerFunc)
 	endpoint.POST("/filter/saved", c.filterSavedAddHandlerFunc)
+	endpoint.POST("/context/table-interval", c.getTableAndIntervalHandlerFunc)
```
vincentbernat, line 102: `/graph/table-interval` is better I think. And, please, add a test for this endpoint.

vincentbernat (changes requested):

vincentbernat:
Sorry, I was too fast to merge. I have added 4eaaf977b77e.

Issue #475, opened 2023-02-12
Make the main "flows" table more space-efficient


vincentbernat:
Compression is not very efficient on this table because its primary key is starting with `TimeReceived`. We should change it to `toStartOfFiveMinutes(TimeReceived)` instead. Like:

```sql
PRIMARY KEY (toStartOfFiveMinutes(TimeReceived), ExporterAddress, InIfName, OutIfName)
```

This needs to be tested by creating a temporary table and use `INSERT SELECT`.

To change the primary key, one can use this: https://kb.altinity.com/altinity-kb-schema-design/change-order-by/

PR #1137, opened 2024-03-30
orchestrator/clickhouse: add support for distributed/replicated tables

### LINKED ISSUE #605: Support for ReplicatedMergeTree in ClickHouse

vincentbernat:
Fix #605

This is only the first step (and not really tested). When enabling "cluster mode", merge tree tables are turned into replicated merge tree tables and we also add distributed version of all tables. Maybe it would be easier to use `_local` prefix instead of `_all` prefix. This way, no need to change anything else (or almost).


vincentbernat:
This has been updated to use `_local` suffix for local tables. Distributed tables keep the same name as in the non-cluster mode. Commit f1fa5bc13f25b9dc2872d2aec1f6df2c8d5ed2b1 contains the data flow between local and distributed tables.

vincentbernat:
While not related, also append a commit to change the primary key for the `flows` table. This would fix #475. This is only effective for new installations until something like ClickHouse/ClickHouse#28065 is implemented.

sn530wdc:
Hi Vincent! We're very interested in this PR. But we are began this activity already on our test-lab. This lab configured without any sampling enabled (for audit purposes, netflow streams are sent to other collectors).
We installed clickhouse on 6 servers, created 6 shards without replica. Kafka created in KRaft mode with 3 nodes, topic have 72 partitions (and every CH server consumes with 12 consumers).
Before running orchestrator with console, we created the following tables with ON CLUSTER statement:
rpl_flows: ReplicatedMergeTree table
flows: Distributed table with rand() inserting key.
rpl_flows_1m0s: ReplicatedMergeTree table.
flows_1m0s: Distributed table with rand() inserting key.
rpl_flows_5m0s: ReplicatedMergeTree table.
flows_5m0s: Distributed table with rand() inserting key.
And, we've encountered some problems during this setup. The main ones are as follows:
1. Visualizing last 6 hours with only one dimension "ExporterName" lasts about 3 minutes (after we modified tables in second problem below).
2. Adding to dimension one of or all of high-cardinalited fields (Src/Dst Addr/Port) throwing a 500 Internal Server Error in a browser, in the logs this is "Unknown column name: <column>". So we're created those ones in aggregated tables (flows_1m0s, flows_5m0s). I know that this caused for long-running selects, but we need to get top charts by addresses and ports, and we are trying to find a middle ground between detailed information and fast query results.
A little more about our loads:
  - 200k flows/s (sampling is disabled, 1)
  - 24 hour gathered data takes up to 100Gb data in kafka, retention is 12 hour in topic configuration
  - In clickhouse, aggregated table takes approximately 80Gb, there is no matter this is 1m0s or 5m0s.

So, can you give us some advices to tune this setup, or maybe wee can be a beta testers of this PR?

vincentbernat:
This PR is merged, so you can already test the result. I don't think this will be much different from your setup.

PR #1167, opened 2024-04-10
Add support for postgresql for filter save

netixx:
Support postgresql as driver in addition to sqlite3.

I opted to use testcontainers to instanciante the database for the tests, let me know if you want to do otherwise.


vincentbernat:
For testing, I'll prefer you use the existing infrastructure: `docker/docker-compose-dev.yml` and `helpers.CheckExternalService()`. I didn't know about testcontainers, it may be interesting to use for tests, but the docker compose stuff also enables local development.

console/database/saved_filters_test.go
```diff
@@ -98,6 +99,58 @@ func TestSavedFilter(t *testing.T) {
...
+    -- replace current_schema() in query with 'schematodeletetablesfrom'
+    -- *and* update the generate 'DROP...' accordingly.
+    FOR r IN (SELECT tablename FROM pg_tables WHERE schemaname = current_schema()) LOOP
+        EXECUTE 'DROP TABLE IF EXISTS ' || quote_ident(r.tablename) || ' CASCADE';
+    END LOOP;
+END $$;
+`); err != nil {
+		t.Fatalf("failed to clean database: %s", err)
+	}
+}
```
vincentbernat, line 152: You should put that as a function called from `t.Cleanup()` to ensure it is still  triggered if there is a test failure.

docker/versions.yml
```diff
@@ -13,7 +13,8 @@ services:
     image: traefik:2.6
   akvorado:
     image: ghcr.io/akvorado/akvorado:main
-
+  postgres:
+    image: docker.io/postgres:16-alpine
```
vincentbernat, line 17: Nitpick: move it in the "for tests" section.

vincentbernat (changes requested):

PR #1257, opened 2024-06-05
Resolve L4 ports protocol names

citizen8:
This PR adds the resolve L4 ports protocol names, example "443:https".
This was asked in the discussions #1166


vincentbernat:
Thanks!

PR #1482, opened 2024-11-05
feat: configure limit type

dimbdr:
Add a limit type combo box, with 2 options:
- Avg -> default type, which works as usual
- Max -> aggregates data focusing on the max value of each axis

Updated documentation

### DIFF
--- a/console/config.go
+++ b/console/config.go
@@ -46,7 +46,7 @@ type VisualizeOptionsConfiguration struct {
 	// Limit is the default limit to use
 	Limit int `json:"limit" validate:"min=5"`
 	// LimitType is the default limitType to use
-	LimitType string `json:"limitType"`
+	LimitType string `json:"limitType" validate:"oneof=Avg Max"`
 	// Bidirectional tells if a graph should be bidirectional (all except sankey)
 	Bidirectional bool `json:"bidirectional"`
 	// PreviousPeriod tells if a graph should display the previous period (for stacked)
--- a/console/config.go
+++ b/console/config.go
@@ -45,6 +45,8 @@ type VisualizeOptionsConfiguration struct {
 	Dimensions []query.Column `json:"dimensions"`
 	// Limit is the default limit to use
 	Limit int `json:"limit" validate:"min=5"`
+	// LimitType is the default limitType to use
+	LimitType string `json:"limitType"`
 	// Bidirectional tells if a graph should be bidirectional (all except sankey)
 	Bidirectional bool `json:"bidirectional"`
 	// PreviousPeriod tells if a graph should display the previous period (for stacked)
@@ -61,6 +63,7 @@ func DefaultConfiguration() Configuration {
 			Filter:     "InIfBoundary = external",
 			Dimensions: []query.Column{query.NewColumn("SrcAS")},
 			Limit:      10,
+			LimitType:  "Avg",
 		},
 		HomepageTopWidgets:     []string{"src-as", "src-port", "protocol", "src-country", "etype"},
 		DimensionsLimit:        50,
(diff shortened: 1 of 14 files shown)


vincentbernat:
I don't know exactly what you would expect with MAX. As flows are aggregated more and more in the various tables, MAX can be widely different while AVG will stay the same. I suppose you implemented it for a reason, but I fear this would be quite disturbing for many users due to the random side of its effect.

moutyque:
Hello @vincentbernat,

By "MAX can be widely different" do you meant after the aggregation of flows for long terme query ?
If so can we imagine to restrict this MAX for recent flows only ? Like checking against which table we do the query and if the table is not "flows" prompt an error message ?
Thx

vincentbernat:
Yes, there is the aggregation done inside ClickHouse, but also the aggregation done by Netflow/IPFIX. Moreover, I don't understand why you group by the timefilter. What would be the effect of such a thing? Why not just take the max instead of the sum?

I understand your use case, but I need to have a balance between features and maintainability. This code adds a lot of branching. I'll have to maintain that forever.

dimbdr:
I slightly reworked the query, the group by timefilter was indeed unnecessary.
Seems to be more accurate now.

As for the code,I also cleaned up since the query is a bit more easier to write.

console/config.go
```diff
@@ -45,6 +45,8 @@ type VisualizeOptionsConfiguration struct {
 	Dimensions []query.Column `json:"dimensions"`
 	// Limit is the default limit to use
 	Limit int `json:"limit" validate:"min=5"`
+	// LimitType is the default limitType to use
+	LimitType string `json:"limitType"`
```
vincentbernat, line 49: Add validation: `validate:"oneof=avg max"`.

console/graph.go
```diff
@@ -18,8 +18,9 @@ type graphCommonHandlerInput struct {
 	schema         *schema.Component
 	Start          time.Time      `json:"start" binding:"required"`
 	End            time.Time      `json:"end" binding:"required,gtfield=Start"`
-	Dimensions     []query.Column `json:"dimensions"`                          // group by ...
-	Limit          int            `json:"limit" binding:"min=1"`               // limit product of dimensions
+	Dimensions     []query.Column `json:"dimensions"`            // group by ...
+	Limit          int            `json:"limit" binding:"min=1"` // limit product of dimensions
+	LimitType      string         `json:"limitType"`
```
vincentbernat, line 23: Also validation here.

console/line.go
```diff
@@ -339,7 +334,11 @@ func (c *Component) graphLineHandlerFunc(gc *gin.Context) {
 			if rows[axis][jKey][0] == "Other" {
 				return true
 			}
-			return sums[axis][iKey] > sums[axis][jKey]
+			if input.LimitType == "Max" {
+				return getMax(points[axis][iKey]) > getMax(points[axis][jKey])
+			} else {
+				return sums[axis][iKey] > sums[axis][jKey]
+			}
```
vincentbernat, line 341: Build `maxes` like `sums` is built (or the other way around, also use a function to get the sum).

console/line.go
```diff
@@ -423,6 +422,16 @@ func (c *Component) graphLineHandlerFunc(gc *gin.Context) {
...
 
+func getMax(values []int) int {
+	valMax := values[0]
+	for _, v := range values {
+		if v > valMax {
+			valMax = v
+		}
+	}
+	return valMax
+}
```
vincentbernat, line 433: There is also `slices.Max()`.

vincentbernat (commented):
Few minor changes.

console/line.go
dimbdr: Removed the `getMax` function, no longer useful

console/line.go
dimbdr: Done, `maxes` is constructed the same way `sums` is

console/graph.go
dimbdr: Done

console/config.go
dimbdr: Done

vincentbernat:
Thanks!

Issue #1311, opened 2024-07-15
Feature request: HeatMap diagrams


yu-re-ka:
as seen in [NetMeta](https://github.com/monogon-dev/NetMeta/):
<img src=https://user-images.githubusercontent.com/859697/103421113-10df0f00-4b9b-11eb-8747-7ccd9c8af76c.png>

PR #1493, opened 2024-11-11
Add Heatmap visualization

yu-re-ka:
Implements #1311 

foo
foo

Todo:

- [x] improve the formatting of the values
- [x] improve the time axis legend


vincentbernat (commented):
Feature-wise, it looks OK, but I don't like it makes the `DataGraphLine.vue` quite harder to read. I think we should move heatmap specific stuff to `DataGraphHeatmap.vue`. To avoid duplication of code, we can factor the common bits into a `useTimeSeriesGraph` (that would go in a file of the same name).

Now that I have described it, I can easily do the changes myself, maybe it would be simpler. Is that OK for you?

PR #1569, opened 2024-12-17
Fix system reload request to include db name

netixx:
Depending on (I guess) database configuration,
it is required to specify the full dict name (include database), or else we get "BAD_ARGUMENTS" error.

It seems like a good idea to always send the full table/dict name regardless, because it makes sure there is no conflict in case of duplicate table names (eg. multiple akvorados running on the same clickhous cluster).


vincentbernat:
Thanks!

PR #1635, opened 2025-02-04
feat/console: DataTable show last value

fabianbees:
# What this feature does:
This is a very small and simple feature, which also shows the last value of the currently selected Time range in the DataTable.

See screenshot:

Bildschirmfoto vom 2025-02-04 08-59-29 (Kopie)

# Why is this useful:
As the last value is mostly of high interest (especially if the End of the Time range is `now`), showing this value also in the DataTable does not require the user to move the cursor to the last value in the Graph. On refreshes the last value can be seen directly in the table and no mouse input is required.

Additionally, if the Graph type `Grid` is used, the last value can't be retrieved easily without this feature:

Bildschirmfoto vom 2025-02-04 09-29-48


vincentbernat:
Thanks!

PR #1645, opened 2025-02-11
Enable clickhouse configs override for console in orchestrator

Itaah:
This PR aims to enable some fields override for ClickHouse configuration in Console configuration, served by orchestrator.
The main goal is to use a Read-only user in ClickHouse for queries, and thus have different profiles/quotas for each usage.


vincentbernat:
I have implemented this differently in 8686e1cbd233, notably to avoid forgetting to update the orchestrator if we accept new keys for ClickHouse configuration. Unlike your proposition, you have to override the whole configuration.

PR #1657, opened 2025-02-21
feat/console: option to sort by last column

fabianbees:
This is a quick followup on my previous PR #1635, which added the `Last` column to the DataTable.

This PR aims to make the DataTable also sortable by the `Last` column, like it is already possible with `Max`and `Avg`.

### Here is a screenshot:

Bildschirmfoto vom 2025-02-21 09-25-50


vincentbernat:
Thanks!

PR #1701, opened 2025-03-19
routing: include routeAttributes.plen in .Equal() and .Hash()

bogi788:
We noticed invalid DstNetMask and SrcNetMask values from routers with only a default route, where the value from flows is zero and replaced with information from the BMP provider.

We were able to confirm that the correct routes were advertised via BMP, and further discovered that the `plen` set in [`inlet/routing/provider/bmp/events.go` line 397](https://github.com/akvorado/akvorado/blob/98922dda9ff05d6d4b4c30781393250c2b353296/inlet/routing/provider/bmp/events.go#L397) did not always match the value when read in [`addPrefix()`](https://github.com/akvorado/akvorado/blob/98922dda9ff05d6d4b4c30781393250c2b353296/inlet/routing/provider/bmp/rib.go#L139).

We believe the cause might be that the `plen` attribute in is not considered in the `Equals()` and `Hash()` functions for the `routeAttribute` struct. This PR includes a fix for that, which seems to solve our issues.

### DIFF
--- a/inlet/routing/provider/bmp/rib.go
+++ b/inlet/routing/provider/bmp/rib.go
@@ -94,6 +94,7 @@ func (rta routeAttributes) Hash() uint64 {
 	if len(rta.communities) > 0 {
 		state.Add((*byte)(unsafe.Pointer(&rta.communities[0])), len(rta.communities)*int(unsafe.Sizeof(rta.communities[0])))
 	}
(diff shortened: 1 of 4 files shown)


vincentbernat:
Thanks! Do you think you can add a test to avoid a future regression?

bogi788:
I've extended the routing component tests with a check that without the changes and passes with them.

vincentbernat:
Thanks!

PR #1702, opened 2025-03-19
static metadata provider: add option skip missing interfaces

bogi788:
As briefly discussed in https://github.com/akvorado/akvorado/discussions/1572 we've implemented a PR for processing only the configured interfaces in the static metadata provider and skip to the next provider for the rest. This behaviour can be enabled per exporter by setting `skip_missing_interfaces` to `true` in the static metadata provider configuration.

For this the `Provider` interface's `Query()` function is updated to use a pointer to the `query` parameter, so that the `ifIndexes` attribute can be updated. The static provider removes interfaces from this slice as it processes them, and returns `ErrSkipProvider` if interfaces were skipped.

### DIFF
--- a/console/data/docs/02-configuration.md
+++ b/console/data/docs/02-configuration.md
@@ -492,6 +492,9 @@ an exporter configuration. An exporter configuration is map:
 - `name` is the name of the exporter
 - `default` is the default interface when no match is found
 - `ifindexes` is a map from interface indexes to interface
+- `skip_missing_interfaces` defines whether the exporter should process only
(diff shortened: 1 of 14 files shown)


bogi788:
We've updated the PR to pass the `BatchRequest` as pointer to make it clear that `ifIndexes` might be modified. I believe this also would be the case when returning new `BatchRequest` objects, unless we allocated a new slice using `make()`/`copy()` before writing to if.

vincentbernat:
Thanks!

PR #1755, opened 2025-05-06
Add depends_on: kafka to avoid crash loop on startup

azery12356:
Wait for kafka to be healthy to start akvorado-orchestrator to avoid a crash loop and service unhealthy when kafka takes too long to start.


vincentbernat:
Thanks!

PR #1847, opened 2025-07-11
inlet: split inlet into new inlet and outlet

### LINKED ISSUE #1122: Kafka consumers wont ever exceed 16
This can be due to the fact clickhouse wont spawn more threads than there is logical cores, but there's also a thread limit on clickhouse that's enforced with other various settings which are defaulted to 16 everywhere in clickhouse.

It seems the most common way to bypass the limitation is to set kafka_disable_num_consumers_limit to 1.

Would worth documenting this and providing a way to turn this on, as it can be quite limiting.

vincentbernat:
This change split the inlet component into a simpler inlet and a new
outlet component. The new inlet component receive flows and put them in
Kafka, unparsed. The outlet component takes them from Kafka and resume
the processing from here (flow parsing, enrichment) and puts them in
ClickHouse.

The main goal is to ensure the inlet does a minimal work to not be late
when processing packets (and restart faster). It also brings some
simplification as the number of knobs to tune everything is reduced: for
inlet, we only need to tune the queue size for UDP, the number of
workers and a few Kafka parameters; for outlet, we need to tune a few
Kafka parameters, the number of workers and a few ClickHouse parameters.

The outlet component features a simple Kafka input component. The core
component becomes just a callback function. There is also a new
ClickHouse component to push data to ClickHouse using the low-level
ch-go library with batch inserts.

This processing has an impact on the internal representation of a
FlowMessage. Previously, it was tailored to dynamically build the
protobuf message to be put in Kafka. Now, it builds the batch request to
be sent to ClickHouse. This makes the FlowMessage structure hides the
content of the next batch request and therefore, it should be reused.
This also changes the way we decode flows as they don't output
FlowMessage anymore, they reuse one that is provided to each worker.

The ClickHouse tables are slightly updated. Instead of using Kafka
engine, the Null engine is used instead.

Fix #1122

## TODO
- [x] fix issue when inserting into ClickHouse
- [x] replace `protoc` by `buf`
- [x] use vtprotobuf
- [x] switch to franz-go (sarama is using channels/goroutines, making it difficult to be more efficient by using "share nothing" workers)
- [x] make metadata query blocking (in the outlet, we have more time to process flows)
- [ ] ~~outlet/clickhouse should also ack what is received by Kafka~~ (not that easy)
- [x] update documentation, notably around scaling


vincentbernat:
Still not working as expected:

```
{"level":"error","worker":4,"error":"send input: write block: encode: raw block: \"EType\" has 7 rows, expected 2","time":"2025-07-12T16:50:46Z","caller":"akvorado/outlet/clickhouse/worker.go:90","module":"akvorado/outlet/clickhouse","message":"cannot send batch to ClickHouse"}
```

vincentbernat:
I think this is over. I plan to merge this soon. I also plan to release a 2.0.0-beta.1.

PR #1920, opened 2025-08-27
outlet: implement RFC 5103 support

### LINKED ISSUE #1827: RFC 5103 “Bidirectional Flow Export” Support
### Disclaimer

- [ ] I plan to implement the feature myself.

### Akvorado version

akvorado v1.11.5

### Feature description

See https://datatracker.ietf.org/doc/html/rfc5103 .  It's used by [ipfixprobe](https://github.com/CESNET/ipfixprobe) by default.

Currently, fields with enterprise bit are ignored:

https://github.com/akvorado/akvorado/blob/b8745754914498ce7e89e0d41f7893d7aa8acaca/inlet/flow/decoder/netflow/decode.go#L118-L120

An implementation could extract the two Uniflows from the Biflow and add them to the protobuf.  Extracting the first Uniflow (forward) is just stripping the reverse fields, extracting the second Uniflow (reverse) would involve stripping the regular (forward) fields and reversing all the “Directional Key Field”s, e.g. sourceIPv4Address.

FestplattenSchnitzel:
Fixes #1827


vincentbernat:
Do you want some help to implement this?

FestplattenSchnitzel:
Your suggestions is doable but I fear the loop will get very complex.

First, a few more if-else branches in the following fashion will be necessary for the key fields, which I think is fine:

```
// L3
case netflow.IPFIX_FIELD_sourceIPv4Address:
	if !isAllZeroIP(v) {
		etype = helpers.ETypeIPv4
		if dir == directionForward {
			bf.SrcAddr = decodeIPFromBytes(v)
		} else {
			bf.DstAddr = decodeIPFromBytes(v)
		}
	}
```

> This should also be more efficient (I think) when you have bidirectional flows as you don't need to merge stuff.

Currently, merges are done for two reasons. First, to make the reverse FlowMessage data available to the orginal FlowMessage which is set up to send to Clickhouse.
Second, to copy values from forward fields which may have been present as reverse field, but weren't.

The first case won't be necessary with your approach.
However the second case currently benefits from having both forward and reverse value of a field available at the same time.
When doing two iterations there needs to be a different solution.
One option would be a struct defined outside of the direction loop storing the forward values of the affected field.
Another would be iterating all relevant fields a third time to fill the gaps.
(I prefer this one.)

An alternative to the two-pass method: take note if a reverse field has been spotted and pass the parameters to a function similar to `decodeRecord` that handles extraction of the reverse flow only.
This would lead to some code duplication though.

I'll go ahead implementing your suggestion and reach out in case I encounter more problems.

> Do you want some help to implement this?

Maybe you have a better idea for solving the problem mentioned above? :)

vincentbernat:
I have added a unittest. It currently does not pass. I will implement my suggestions next.

PR #1932, opened 2025-09-02
common/embed: replace all go:embed use by an embedded archive

vincentbernat:
Some of the files were quite big:

- asns.csv ~ 3 MB
- index.js ~ 1.5 MB
- *.svg ~ 2 MB

Use a ZIP archive to put them all and embed it. This reduce the binary size from 89 MB to 82 MB. 🤯

Not sure to keep that. We'll see.


vincentbernat:
Merged manually.

PR #2058, opened 2025-10-31
Add Netbox API as example to network-sources

sorano:


sorano:
> Thanks, could you remove the TLS section? Since it uses the default values...

Thank you aswell for your help.

I've removed default TLS values.

PR #2075, opened 2025-11-04
Switch to GoBGP v4

vincentbernat:
See https://github.com/osrg/gobgp/issues/3043.

The main changes are:
- usage of `netip.Addr` and `netip.Prefix`
- `bgp.RouteFamily` is now `bgp.Family`
- `AddrPrefixInterface` becomes `PathNLRI`, which splits out path identifier in `ID` and actual NLRI in `NLRI`
- `NLRI` are now v4/v6 independant
- `AFI()`/`SAFI()` methods are not present anymore in `NLRI`


vincentbernat:
The build failure on MacOS is puzzling, but it seems unrelated to this PR.

PR #2109, opened 2025-11-18
orchestrator: add manage-topic flag and document skip-migrations

rastsislau-matusevich:
- Add ManageTopic flag to orchestrator/kafka configuration (default: true)
- Skip Kafka initialization when ManageTopic is false
- Add tests for ManageTopic configuration
- Document manage-topic flag in Kafka section
- Document skip-migrations flag in ClickHouse section (documentation was missing)

This allows running multiple orchestrator instances where only one manages the Kafka topic, useful for multi-datacenter deployments.

Background:

We are planning to add ability to store the flow data both locally and in centralized cloud storage.
To ensure scalability and reliability, we plan to have the following architecture:

Within datacenter, we'd have "main" orchestrator (manages local Kafka + local CH schema) that manages inlet, kafka queue and outlet that would write to a local clickhouse.
Additionally, we'd have "remote-purpose" orchestrator (doesn't manage anything) that manages outlet that would write to remote destination.

Within cloud, we'd have a "remote" orchestrator (doesn't manage any Kafka, manages remote CH schema) that is just responsible for managing remote CH schema and centralized UI.

Adding ability to not manage Kafka would make this setup simpler. There's already a way to avoid managing CH schema, but it was undocumented.

Tested by deploying the modified code and verifying it works (both with flag enabled and disabled)

### DIFF
--- a/cmd/orchestrator.go
+++ b/cmd/orchestrator.go
@@ -233,9 +233,7 @@ func orchestratorStart(r *reporter.Reporter, config OrchestratorConfiguration, d
 		httpComponent,
 		clickhouseDBComponent,
 		clickhouseComponent,
-	}
-	if kafkaComponent != nil {
-		components = append(components, kafkaComponent)
+		kafkaComponent,
 	}
 	return StartStopComponents(r, daemonComponent, components)
 }


orchestrator/kafka/config_test.go
```diff
@@ -7,12 +7,36 @@ import (
...
+		t.Fatalf("New() error:\n%+v", err)
+	}
+	if c.kafkaOpts != nil {
+		t.Error("kafkaOpts should be nil when ManageTopic is false")
+	}
+	if c.kafkaTopic != "" {
+		t.Error("kafkaTopic should be empty when ManageTopic is false")
+	}
+	if err := c.Start(); err != nil {
+		t.Fatalf("Start() error:\n%+v", err)
```
vincentbernat: Use `helpers.StartStop(t, c)`.

orchestrator/kafka/root.go
```diff
@@ -36,19 +36,25 @@ type Dependencies struct {
...
-
 	c := Component{
 		r:      r,
 		d:      dependencies,
 		config: config,
+	}
+
+	if !config.ManageTopic {
+		r.Info().Msg("Kafka topic management disabled, skipping Kafka initialization")
+		return &c, nil
```
vincentbernat: I would suggest to just return `nil, nil` early and handle the nil case in `Start()` and `Stop()`.

vincentbernat (changes requested):

cmd/orchestrator.go
```diff
@@ -233,7 +233,9 @@ func orchestratorStart(r *reporter.Reporter, config OrchestratorConfiguration, d
 		httpComponent,
 		clickhouseDBComponent,
 		clickhouseComponent,
-		kafkaComponent,
+	}
+	if kafkaComponent != nil {
+		components = append(components, kafkaComponent)
```
vincentbernat: I don't think this is needed. A component can be nil, as long as Start()/Stop() can handle it.

PR #2112, opened 2025-11-22
Don't disable DCE

vincentbernat:
Akvorado binaries are larger than they should be because dead code elimination is disabled because of using `reflect.MethodByName` or `reflect.Method`. This needs to be fixed everywhere.

- [x] disable tracing support in gRPC (5ae5d966437964e2224b4b973451f07d05ae815a)
- [x] replace static usage of `reflect.MethodByName()` with an interface check (756d0a958012d4e5390aaefc0e17cd56cc352b33) 
- [x] fix [go-playground/validator/v10](https://github.com/go-playground/validator/issues/1496): 00a9ecd36635
- [x] don't use `text/template`
- [ ] fix [expr-lang/expr](https://github.com/expr-lang/expr/pull/956)
- [x] fix [gorm.io/gorm](https://github.com/go-gorm/gorm/pull/7643), removed in 8e2b97aa0257f87f91b489d0a3fbb57feae58041
- [x] add a non-regression test

### DIFF
--- a/cmd/akvorado/demo-exporter.go
+++ b/cmd/akvorado/demo-exporter.go
@@ -39,6 +39,11 @@ func (c *DemoExporterConfiguration) Reset() {
 	}
 }
 
+// UnusedButExportedFunction is there to test dead code elimination.
+func (c *DemoExporterConfiguration) UnusedButExportedFunction() {
+	panic("never call me!")
+}
+
 type demoExporterOptions struct {
 	ConfigRelatedOptions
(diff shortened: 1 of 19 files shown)


vincentbernat:
I did not manage to resolve conflicts correctly. It does not matter as this is still not done.

vincentbernat:
Alternatively, instead of templating, #2389 could solve a big chunk of the templating.

PR #2158, opened 2025-12-09
console: use URI-safe encoding for state

frazew:
# Summary

"/" is by default in the base64 charset, Vue(?) URL-encodes it to %2F but the HTTP spec expects / and %2F to be treated the same way. Curiously, traefik 3.6.4 decided to break this behavior and rejects URLs with %2F.

lz-string provides a URI-safe encoding scheme (which just maps base64 to a URI-safe charset), let's use that by default and try to decode from base64 otherwise to preserve backwards-compatibility.

# Notes

This is caused by https://github.com/traefik/traefik/pull/12360 (traefik 3.6.4).

Example (invalid JSON but it's the bare minimum) string which reproduces the issue:
```typescript
import LZString from "lz-string";
let encoded = LZString.compressToBase64('{"bidirectional":false,"dimensions":["DstAddr","SrcAddr"],"end":"2025-12-05T15:23:40.562Z","filter":"OutIfBo');
console.log(encoded);
// N4IgRglgJhBOCmBjALhA9gOwIYBsQC4AzXAZ3gBoQYBbeDE9eggbRABETkBBKKWESgGVYiHnxABdSnSgEQAJgAM8gKwBaAIzy1ilQBUNK/PIDM+ACyKAdCoBs8gFoCQhCDmTx++EAHkArsgAkoQAQmhAA===
// -> the output string contains a "/" which gets URL-encoded to %2F and is then rejected by the Go http URI parser
```

When the user refreshes the page, traefik receives the `%2F` and rejects it:
```
akvorado-traefik-1  | 2025-12-09T10:45:01Z DBG github.com/traefik/traefik/v3/pkg/server/server_entrypoint_tcp.go:811 > Rejecting request because it contains encoded character %2F in the URL path: /visualize/N4IgRglgJhBOCmBjALhA9gOwIYBsQC4AzXAZ3gBoQYBbeDE9eggbRABETkBBKKWESgGVYiHnxABdSnSgEQAJgAM8gKwBaAIzy1ilQBUNK%2FPIDM+ACyKAdCoBs8gFoCQhCDmTx++EAHkArsgAkoQAQmhAA===
```

As a side note, should this be in the fragment section of the URI instead of the path? To simply avoid sending it to the server


frazew:
I initially also wondered if an "old version" encoding could pass the "new version" decoding but result in a messy string and my conclusion was that is was most likely fine because the charsets are the same except for the two last characters (see [base64 here](https://github.com/pieroxy/lz-string/blob/9a3d312847ed13d394a908adc03dbdf9e62d44d8/src/base64/keyStrBase64.ts) and [uriSafe here](https://github.com/pieroxy/lz-string/blob/9a3d312847ed13d394a908adc03dbdf9e62d44d8/src/encodedURIComponent/keyStrUriSafe.ts)). So an "old" base64 string that happened not to use the last 2 chars should decode correctly with uriSafe, and if it uses the last 2 chars then decoding simply fails (because the char is not in the charset).

But honestly I also agree that prepending a "version" identifier is probably safer because it means the behavior doesn't rely on the internal implementation of the library, I'll update the patch :smile:

vincentbernat (approved):

PR #2211, opened 2026-01-15
allow to configure clickhouse storage policy for resolution tables

### LINKED ISSUE #2210: Allow to set a clickhouse storage policy for flows tables
### Disclaimer

- [x] I plan to implement the feature myself.

### Akvorado version

v2

### Feature description

We would like to store the largest resolution of flow-data on a different storage then everything else. Therefor it would be nessesary to have the option to configure a storage policy on the different resolutions.

I would propose the addition of a configuration option for the resolutions which would then get added to the Table-Settings (e.g. `ALTER TABLE flows_1h0m0s MODIFY SETTING storage_policy='s3'`)

n0emis:
Closes #2210

### DIFF
--- a/console/data/docs/02-configuration.md
+++ b/console/data/docs/02-configuration.md
@@ -939,15 +939,16 @@ provided inside `clickhouse`:
   schema mismatches may cause write errors.
 
 The `resolutions` setting contains a list of resolutions. Each resolution has
-two keys: `interval` and `ttl`. The first one is the consolidation interval. The
-second is how long to keep the data in the database. If `ttl` is 0, then the
-data is kept forever. If `interval` is 0, it applies to the raw data (the one in
-the `flows` table). For each resolution, a materialized view `flows_DDDD` is
-created with the specified interval. It should be noted that consolidated tables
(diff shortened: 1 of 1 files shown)


vincentbernat:
Maybe we should provide something more generic, like `TableSettings`, a map from settings names to values. Merge this map into a map containing `index_granularity` = 8192 and `ttl_only_drop_parts` = 1. You need to sort the keys for the output to be stable.

orchestrator/clickhouse/migrations_helpers.go
```diff
@@ -610,7 +613,7 @@ outer:
 	}
 
 	// Check if we need to update the settings
-	settingsClauseLike := fmt.Sprintf("CAST(engine_full LIKE '%% SETTINGS %s', 'String')", settings)
+	settingsClauseLike := fmt.Sprintf("CAST(engine_full LIKE '%% SETTINGS %s', 'String')", strings.ReplaceAll(settings, "'", "\\'"))
```
vincentbernat: Also escape `\`.

PR #2244, opened 2026-02-17
outlet/routing: read from RIB without any locks

### LINKED ISSUE #2092: Reduce lock contention when receiving a lot of BMP messages
When receiving a lot of BMP messages, we have a giant lock protection the routing table. This makes `Lookup()` wait on the lock. See:

Image

BART has lock-free methods, but this is only one part of the solution, as we have a map and the intern values, both of them requiring a lock. The intern values could be replaced by Go's `unique` package (or just stealing its way of keeping references). For the map, `sync.Map` could be used.

Also see #254, #255, #278, and #1461 for some of the previous attempts. Prioritizing readers is not a good solution either, as the writer may be starved and declared dead by the remote BMP sender (this is #1461).

The discussion below contains more profiling graphs.

### Discussed in https://github.com/akvorado/akvorado/discussions/2090

<div type='discussions-op-text'>

<sup>Originally posted by **bogi788** November 11, 2025</sup>
Since updating our Akvorado setup to 2.0 we're having issues with the outlet not processing incoming flows fast enough (and the Kafka consumer group lag increasing) with too many BMP messages arriving. Temporarily blocking BMP (e.g. using an iptables rule) allows the outlet to catch up.

Are there any workaround we can apply (besides using two orchestrators and outlets writing to the same clickhouse database)?</div>

vincentbernat:
This series modifies the BMP provider for the routing component by removing locks.

The intern package is replaced by Go's unique package. This is more efficient than trying to make the intern package thread-safe. Lookup is not impacted by the change, insertion is slower, but routes take less space and flushing is faster (compared to the thread-safe version of the intern package).

bart's persist tables are used to make the RIB thread-safe. This has a huge performance impact on insertion performance (×10), as well as space (×2). Impact on lookup is small (~ 7%).

The next step is to make paths thread-safe by using some kind of sync.Map. The last commit swith to xsync.Map which seems the most efficient for our use case.

At the end, 42% increase in space, 1000% increase in insertion time, 12% lookup. The benchmarks do not test performance of multiple routines updating/looking up the RIB, but the idea is that it should behave better without locks, even if it is slower during updates. So, hopefully, someone could test the result for real.

Fix #2092.

### DIFF
--- a/bench-step0.txt
+++ b/bench-step0.txt
@@ -1,0 +1,215 @@
+[34;1m▶[0m formatting Go code…
+[34;1m▶[0m running golint…
+[34;1m▶[0m compiling protocol buffers common/pb/rawflow.pb.go…
+[34;1m▶[0m generate column definitions…
+[34;1m▶[0m generate PEG parser for filters…
+[34;1m▶[0m generate enums for LoadBalanceAlgorithm…
+[34;1m▶[0m generate enums for ASNProvider…
+[34;1m▶[0m generate enums for NetProvider…
+[34;1m▶[0m generate enums for AuthProtocol…
+[34;1m▶[0m generate enums for PrivProtocol…
+[34;1m▶[0m generate enums for IfSpeedPathUnit…
+[34;1m▶[0m generate enums for HomepageTopWidget…
+[34;1m▶[0m generate enums for SASLMechanism…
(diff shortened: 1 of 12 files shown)


vincentbernat:
See tests in #2092. This does not help.

PR #2269, opened 2026-02-24
Draft: prepare data before locking in handleRouteMonitoring

bogi788:
This PR reorders the steps in `handleRouteMonitoring` for the BMP routing provider in order to reduce the time the routing table is locked.

It achieves this by performing most operations that don't access the routing table first, storing intermediate results in dynamic arrays.

### DIFF
--- a/outlet/routing/provider/bmp/events.go
+++ b/outlet/routing/provider/bmp/events.go
@@ -232,24 +232,14 @@ func (p *Provider) handleRouteMonitoring(pkey peerKey, body *bmp.BMPRouteMonitor
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


vincentbernat:
It makes sense, but this is a trade-off: arrays put a burden on the GC. Do you notice an improvement on your side after applying this?

bogi788:
I've only compared with different exporters with comparable number of bmp messages.

I could only notice a difference after restarting the outlet (or bmp on the routers):

without this PR:
profile001

with this PR:
profile002

But there basically no difference once BMP has "settled" and only updates are sent:

without:
profile005

with:
profile006

PR #2294, opened 2026-03-05
outlet/routing: implement RIB sharding for BMP

vincentbernat:
This enables more finer-grained locks and should help with lock contention with almost no downside (except increased memory usage).

It was not benchmarked yet.

Use of `struct rawRoute` could be reversed by providing a function to get a RIB shard from a prefix and have `AddRoute()` works on that. This way, we can avoid internalising several times the same variable.

### DIFF
--- a/cmd/akvorado/testdata/configurations/bmp-to-routing/expected.yaml
+++ b/cmd/akvorado/testdata/configurations/bmp-to-routing/expected.yaml
@@ -11,6 +11,7 @@ paths:
       keep: 1h0m0s
       rds: []
       receivebuffer: 0
+      ribshards: 16
   outlet.0.core.asnproviders:
     - flow
     - routing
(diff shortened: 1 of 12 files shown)


vincentbernat:
@bogi788 I did not finish yet, but you may want to try it to see if it improves. Remaining:

- benchmark
- add a method `WithShard(pfx netip.Prefix, func())` to rib that would invoke the callback with the shard for the prefix and the lock held. This should allow interning less values (notably nexthop and route attributes can be interned once for several NLRIs).

vincentbernat:
> I've switched the outlet to this branch ([505f9e3](https://github.com/akvorado/akvorado/pull/2294/commits/505f9e3157741333500ab88366c0a6d5a5b59645) from earlier) for testing. It is processing flows around 50% faster than the main branch after the same uptime. I haven't noticed a big difference in CPU utilisation yet.

Oh, that's good. Is memory usage different? You could also try to tune the number of shards. Like 8, 16 (default), 32, 64, to see if it's better with more or less.

bogi788:
RAM usage on the VM seems unchanged, but I can't see individual processes or containers in our metrics.

PR #2319, opened 2026-03-13
Support skip missing interfaces for remote metadata sources

adriansmares:
References https://github.com/akvorado/akvorado/discussions/2318

This MR adds support for `skipmissinginterfaces` to remote metadata sources.

### DIFF
--- a/console/data/docs/02-configuration.md
+++ b/console/data/docs/02-configuration.md
@@ -494,7 +494,7 @@ sources. Each source accepts these attributes:
   `.[1:][]`. The `plain` parser produces an array of strings, one per line.
 - `transform` is a [jq](https://stedolan.github.io/jq/manual/) expression that
   transforms the parsed data into a set of attributes represented as objects.
-  Each object should have these keys: `exporter-subnet`, `skip-missing-interfaces`,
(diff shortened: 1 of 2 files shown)


outlet/metadata/provider/static/source_test.go
```diff
@@ -235,3 +235,96 @@ func TestRemoteExporterSources(t *testing.T) {
...
+	mux.Handle("/exporters.json", http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Add("Content-Type", "application/json")
+		w.WriteHeader(200)
+		w.Write([]byte(`
+{
+  "exporters": [
+    {
+      "exportersubnet": "2001:db8:1::/48",
+      "name": "remote-exporter",
+      "skipmissinginterfaces": true,
```
vincentbernat: `skip-missing-interfaces`. This way, you test it works as advertised.

vincentbernat (changes requested):
So, rollback the change. But if the tests are OK with `skip-missing-interfaces` instead of `skipmissinginterfaces`, it means the tests are not working as expected?

PR #2368, opened 2026-04-09
orchestrator/clickhouse: Bloom Filter Skipping Index

PolyAnthi:
For extremely large datasets, query times can become an issue, especially for IPs that are not often appearing in data since ClickHouse needs to scan the entire table.

To resolve this, we can use bloom filters, which are natively supported in ClickHouse, and exchange a very small amount of disk space (I personally saw 600M rows for about 160MB at 0.001 FPP) for its indexing. This can be enabled and disabled for each Addr field respectively.

I've seen query times decrease by as much as 98% in some cases where you just want traffic within last 30 days for a specific IP.


vincentbernat:
Interesting. I didn't know about skip indexes in ClickHouse. This brings many questions on my side.

The main table is using `ORDER BY (toStartOfFiveMinutes(TimeReceived), ExporterAddress, InIfName, OutIfName)
`. The main idea of that is that every request use `TimeReceived`. The reasons for others are to get better compression since many other columns derive from these columns. Now I think it would be better to have:

- PRIMARY KEY toStartOfFiveMinutes(TimeReceived)
- ORDER BY (toStartOfFiveMinutes(TimeReceived), ExporterAddress, InIfName, OutIfName)
- bloom filters for SrcAS/DstAS, SrcPort/DstPort, SrcCountry/DstCountry, SrcAddr/DstAddr, ExporterName, In/OutIfProvider, In/OutIfConnectivity, In/OutIfBoundary.

Unfortunately, we cannot change the PRIMARY KEY yet. It would be another migration to do manually (the second one).

Back to your PR, I think it would be better to enable/disable bloom filters in the schema directly, instead of hardcoding for SrcAddr/DstAddr.

```yaml
schema:
  indexes:
    SrcAddr: 0.01
  noindexes:
    - DstAS
```

I am not sure everything should be a bloom filter. For some columns, it may make sense to use minmax (exportername for example since we are ordering by it quite early) or set(n) the low cardinality ones like connectivity, provider, boundaries. So `indexes` could be `minmax`, `set(X)` or `bloom(0.01)`.

As for the default values we could provide, I need to look at [ClickHouse best practices](https://clickhouse.com/docs/best-practices/use-data-skipping-indices-where-appropriate).

PolyAnthi:
Yeah, ClickHouse skip indexes are pretty slept on in my opinion. They give insane benefits.

I will take a look at shifting the config spot for those and seeing if it makes sense to do it on all of those columns, or only some of them etc and determine the right type (or even make it configurable to user, TBD).

As for default values, may be best to just do some very basic testing to get a decent value. The best practices page doesn't seem to provide any values or benchmarks for datasets/FPP values. I think the best case here for default values is using a good bit of common sense to make a reasonable balance then making sure config examples and/or documentation have it explained in a bit more detail for the users who would like to tweak these settings.

PolyAnthi:
@vincentbernat

Test data is all updated
Updates to new methods in very similar style as you requested, except allowing for more types

Default values & Config examples:
https://github.com/PolyAnthi/akvorado/blob/main/console/data/docs/02-configuration.md#data-skipping-indexes

Test actions all passed:
https://github.com/PolyAnthi/akvorado/actions/runs/24296285319

vincentbernat:
Thanks, that was pretty fast! I'll try to review this quickly. The documentation looks good, but you say there is a materialization when setting up an index while the [documentation says otherwise](https://clickhouse.com/docs/optimize/skipping-indexes#:~:text=Normally%20skip%20indexes%20are%20only%20applied%20on%20newly%20inserted%20data%2C%20so%20just%20adding%20the%20index%20won%27t%20affect%20the%20above%20query).

orchestrator/clickhouse/migrations_helpers.go
```diff
@@ -687,12 +693,118 @@ outer:
 		modified = true
 	}
 
+	// Data-skipping indices are applied to the main flows table only, since
+	// some indexed columns (e.g. SrcAddr/DstAddr) are ClickHouseMainOnly.
```
vincentbernat: I think it would be better to apply them on other tables as well, as long as the column exists there. For many of them, it makes more sense to apply them on other tables as they are more likely to be used there.

orchestrator/clickhouse/config_test.go
```diff
@@ -157,6 +157,7 @@ func TestDefaultConfiguration(t *testing.T) {
 	}
 }
 
+
```
vincentbernat: Extra blank line.

orchestrator/clickhouse/migrations_helpers.go
```diff
@@ -687,12 +693,118 @@ outer:
...
+		idxType := skipIndexes[colKey]
+		col, ok := c.d.Schema.LookupColumnByKey(colKey)
+		if !ok || col.Disabled {
+			continue
+		}
+		chType, err := idxType.ClickHouseType()
+		if err != nil {
+			return false, fmt.Errorf("schema index for %s: %w", col.Name, err)
+		}
+		idxName := "idx_" + strings.ToLower(col.Name)
```
vincentbernat:
```suggestion
		idxName := fmt.Sprintf("idx_%s", strings.ToLower(col.Name))
```

common/schema/config.go
```diff
@@ -21,6 +59,13 @@ type Configuration struct {
 	NotMainTableOnly []ColumnKey `validate:"ninterfield=MainTableOnly"`
 	// Materialize lists columns that shall be materialized at ingest instead of computed at query time
 	Materialize []ColumnKey
+	// Indexes maps column names to the desired ClickHouse data-skipping index.
+	// Accepted values: "minmax", "set(N)" (N >= 0), or "bloom(P)" (0 < P < 1).
+	// Indexes are applied only to the main flows table. Entries here are merged
+	// with (and override) the defaults; use NoIndexes to remove a default.
+	Indexes map[ColumnKey]SkipIndexType
+	// NoIndexes lists columns whose default skip index should be removed.
+	NoIndexes []ColumnKey
```
vincentbernat, line 72:
```suggestion
	NoIndexes []ColumnKey `validate:"ninterfield=Indexes"`
```

common/schema/config.go
```diff
@@ -21,6 +59,13 @@ type Configuration struct {
 	NotMainTableOnly []ColumnKey `validate:"ninterfield=MainTableOnly"`
 	// Materialize lists columns that shall be materialized at ingest instead of computed at query time
 	Materialize []ColumnKey
+	// Indexes maps column names to the desired ClickHouse data-skipping index.
+	// Accepted values: "minmax", "set(N)" (N >= 0), or "bloom(P)" (0 < P < 1).
+	// Indexes are applied only to the main flows table. Entries here are merged
+	// with (and override) the defaults; use NoIndexes to remove a default.
+	Indexes map[ColumnKey]SkipIndexType
```
vincentbernat, line 70: You should add tests for validation (if not done). I am not 100% sure, but I think you need:


```suggestion
	Indexes map[ColumnKey]SkipIndexType `validate:"dive"`
```

common/schema/config.go
```diff
@@ -5,10 +5,48 @@ package schema
...
+			return "", fmt.Errorf("invalid bloom index %q: FPP must be in (0, 1)", s)
+		}
+		return fmt.Sprintf("bloom_filter(%g)", p), nil
+	default:
+		return "", fmt.Errorf("unknown skip index type %q: use minmax, set(N), or bloom(P)", s)
+	}
+}
+
+// Validate checks that the SkipIndexType is well-formed.
+func (s SkipIndexType) Validate() error {
```
vincentbernat: In other places, we use `UnmarshalText()` for validation.

common/schema/root.go
```diff
@@ -84,6 +97,19 @@ func New(config Configuration) (*Component, error) {
 		}
 	}
 
+	// Validate skip index types
+	for col, idxType := range config.Indexes {
+		if err := idxType.Validate(); err != nil {
```
vincentbernat: Oh, you use it there. I thought this was a function to be used by validator. So, I think you can skip the suggestion on tagging `dive` for `validate`, but you can keep using `UnmarshalText()` suggestion instead of `Validate()`, since it outputs the error sooner. It does not make much of a difference for the orchestrator, but this is just for consistency with how we validate configuration settings.

config/akvorado.yaml
```diff
@@ -97,6 +99,15 @@ clickhouse:
...
 
+schema:
+  # Sensible data-skipping indexes are enabled by default (see docs/02-configuration.md).
+  # Use `indexes` to add/override and `no-indexes` to remove a default. Example:
+  # indexes:
+  #   SrcAddr: bloom(0.01)   # tighten FPP
+  #   SrcNetName: bloom(0.001)
+  # no-indexes:
+  #   - DstAddr
+
```
vincentbernat: Since this is well documented, I prefer not to bloat the documentation with settings most people won't use.

orchestrator/clickhouse/migrations_helpers.go
```diff
@@ -538,6 +538,12 @@ SETTINGS {{ .Settings }}
 		if err := c.d.ClickHouse.ExecOnCluster(ctx, createQuery); err != nil {
 			return fmt.Errorf("cannot create %s: %w", tableName, err)
 		}
+		// Apply skip indexes to the newly created main table.
+		if resolution.Interval == 0 {
```
vincentbernat: Same comment as below.

orchestrator/clickhouse/migrations_test.go
```diff
@@ -489,6 +489,143 @@ func TestTableSettings(t *testing.T) {
...
+		sch, err := schema.New(cfg)
+		if err != nil {
+			t.Fatalf("schema.New() error:\n%+v", err)
+		}
+		startTestComponentWithConfig(t, r, chComponent, sch, nil)
+		gotMetrics := r.GetMetrics("akvorado_orchestrator_clickhouse_migrations_", "applied_steps_total")
+		if diff := helpers.Diff(gotMetrics, map[string]string{"applied_steps_total": "0"}); diff != "" {
+			t.Fatalf("Metrics (-got, +want):\n%s", diff)
+		}
+	})
```
vincentbernat: This one is already tested as part of the general harness.

orchestrator/clickhouse/migrations_test.go
```diff
@@ -489,6 +489,143 @@ func TestTableSettings(t *testing.T) {
...
+			t.Fatalf("skip index %s: got type %q, want %q", indexName, gotType, wantType)
+		}
+	}
+
+	t.Run("default indexes applied", func(t *testing.T) {
+		checkSkipIndex(t, "idx_srcaddr", "bloom_filter(0.001)")
+		checkSkipIndex(t, "idx_dstaddr", "bloom_filter(0.001)")
+		checkSkipIndex(t, "idx_exportername", "minmax")
+		checkSkipIndex(t, "idx_inifboundary", "set(0)")
+	})
```
vincentbernat: This one is already tested as part of the general harness.

console/data/docs/02-configuration.md
```diff
@@ -780,6 +780,55 @@ For ICMP, you get `ICMPv4Type`, `ICMPv4Code`, `ICMPv6Type`, `ICMPv6Code`,
...
+    - DstAddr
+```
+
+Indexes are idempotent: changing the type or removing an entry will cause
+*Akvorado* to drop and recreate (or drop) the index on the next startup.
+
+> [!NOTE]
+> Adding a skip index to a large table triggers a background materialization
+> step in ClickHouse. Query performance improves gradually as parts are
+> rewritten. Materialization uses disk I/O but does not block ingestion.
```
vincentbernat: So, ClickHouse documentation says otherwise. If true, you could say otherwise that only access to new data is optimized (so people don't expect improvements right now), with maybe the command to trigger materialization.

vincentbernat (changes requested):
A few comments. But otherwise, I like where this is going. Thanks!

PolyAnthi:
> Thanks, that was pretty fast! I'll try to review this quickly. The documentation looks good, but you say there is a materialization when setting up an index while the [documentation says otherwise](https://clickhouse.com/docs/optimize/skipping-indexes#:~:text=Normally%20skip%20indexes%20are%20only%20applied%20on%20newly%20inserted%20data%2C%20so%20just%20adding%20the%20index%20won%27t%20affect%20the%20above%20query).

This was primarily targeted at people who had large amounts of pre-existing data. It will simply just materialize the index to catch up whatever data skipping index is in use.

orchestrator/clickhouse/config_test.go
PolyAnthi: Resolved

common/schema/config.go
PolyAnthi: > So, I think you can skip the suggestion on tagging dive for validate, but you can keep using UnmarshalText() suggestion instead of Validate()

Will skip dive tag and use UT()

console/data/docs/02-configuration.md
PolyAnthi: https://github.com/PolyAnthi/akvorado/blob/268d7f719040c7dbdd14032c0c90607e8a153492/orchestrator/clickhouse/migrations_helpers.go#L765
> "ALTER TABLE %s MATERIALIZE INDEX %s",

By default, ClickHouse doesn't, but we materialize said index upon creation so its able to backfill data to be utilised at query time.

vincentbernat: Understood. I think this is a bad idea. In 15d982165267b5a58177938721c9a3978b65d6e9, I did disable TTL materialization as it can cause operational difficulties on a busy ClickHouse database. I prefer if you document the fact that the effect is not immediate and interested users can still materialize the indexes themselves if they want.

common/schema/config.go
PolyAnthi: The ninterfield validator doesn't work on maps, we can either remove this validation (as it gets handled later anyways) or modify `noIntersectField` in common/helpers/validator.go to accept maps.

What would you prefer?

vincentbernat: Just remove the check. It was a nice to have, but not worth spending time.

console/data/docs/02-configuration.md
PolyAnthi: Resolved https://github.com/akvorado/akvorado/pull/2368/changes/5c2b24b3f112e26e2bed9ab286db523f5ca279c0 and documentation updated in https://github.com/akvorado/akvorado/pull/2368/commits/5d21a5ea45265badcdab36c1af5adead1a2af640

common/schema/config.go
PolyAnthi: https://github.com/akvorado/akvorado/pull/2368/commits/7a2ba60789d14d9fb679e0d2112747d7ddc13d0a

PolyAnthi:
@vincentbernat Currently rerunning actions on https://github.com/PolyAnthi/akvorado/actions/runs/24301316698 - just making sure all tests pass

common/schema/config.go
```diff
@@ -50,6 +99,25 @@ type CustomDictAttribute struct {
...
+	ColumnDstPort:           "bloom(0.001)",
+	ColumnSrcCountry:        "bloom(0.001)",
+	ColumnDstCountry:        "bloom(0.001)",
+	ColumnExporterName:      "minmax",
+	ColumnInIfProvider:      "set(0)",
+	ColumnOutIfProvider:     "set(0)",
+	ColumnInIfConnectivity:  "set(0)",
+	ColumnOutIfConnectivity: "set(0)",
+	ColumnInIfBoundary:      "set(0)",
+	ColumnOutIfBoundary:     "set(0)",
```
vincentbernat, line 118: Maybe not now, but we could tune the default values on a real installation (or at least on `demo.akvorado.net`) and check with `EXPLAIN indexes = 1` if these values make sense (and their sizes).

orchestrator/clickhouse/migrations_helpers.go
```diff
@@ -687,12 +690,114 @@ outer:
...
+		// columns are absent from aggregated tables).
+		var colExists uint64
+		if err := c.d.ClickHouse.QueryRow(ctx,
+			`SELECT count() FROM system.columns WHERE database = $1 AND table = $2 AND name = $3`,
+			c.d.ClickHouse.DatabaseName(), tableName, col.Name).Scan(&colExists); err != nil {
+			return false, fmt.Errorf("cannot check column %s in %s: %w", col.Name, tableName, err)
+		}
+		if colExists == 0 {
+			continue
+		}
```
vincentbernat, line 747: I prefer you check ClickHouseMainOnly. The main reason is that it helps detecting bugs where the tables are deviating from the schema.

orchestrator/clickhouse/migrations_helpers.go
```diff
@@ -687,12 +690,114 @@ outer:
...
+	}
+
+	// Collect stale skip indices that are no longer wanted.
+	rows, err := c.d.ClickHouse.Query(ctx,
+		`SELECT name FROM system.data_skipping_indices WHERE database = $1 AND table = $2 AND startsWith(name, 'idx_')`,
+		c.d.ClickHouse.DatabaseName(), tableName)
+	if err != nil {
+		return false, fmt.Errorf("cannot list skip indices for %s: %w", tableName, err)
+	}
```
vincentbernat: Collect all skip indexes once for both unused indexes and type checking.

vincentbernat (changes requested):
Two small changes. Thanks!

common/schema/config.go
PolyAnthi: Yeah, will need to see how it looks in places with lots of data but I don't see this being an issue. Should be pretty cheap regardless.

orchestrator/clickhouse/migrations_helpers.go
PolyAnthi: https://github.com/akvorado/akvorado/pull/2368/commits/a805669b2aacdc1b2763b45ea7f70fc96a6c0ecc

orchestrator/clickhouse/migrations_helpers.go
PolyAnthi: https://github.com/akvorado/akvorado/pull/2368/commits/f8a030a95618360184df2d7acf1f36a4b0a1b64c

orchestrator/clickhouse/migrations_helpers.go
```diff
@@ -687,12 +690,107 @@ outer:
...
 }
 
+// applySkipIndexes reconciles the skip indexes on tableName with the configured
+// schema indexes. It returns true if any change was made.
+func (c *Component) applySkipIndexes(ctx context.Context, tableName string, isMainTable bool) (bool, error) {
+	skipIndexes := c.d.Schema.GetSkipIndexes()
+	var toDrop []string
+	var toAdd []string
+
+	// Collect all existing skip indexes once for both of the below.
```
vincentbernat:
```suggestion
	// Collect existing skip indexes.
```

Coding agent always like to make comments that only make sense when you have the prompt.

orchestrator/clickhouse/migrations_helpers.go
```diff
@@ -687,12 +690,107 @@ outer:
...
+	existingIndexes := map[string]string{}
+	for rows.Next() {
+		var name, typeFull string
+		if err := rows.Scan(&name, &typeFull); err != nil {
+			rows.Close()
+			return false, fmt.Errorf("cannot scan skip index: %w", err)
+		}
+		existingIndexes[name] = typeFull
+	}
+	rows.Close()
```
vincentbernat: Use `defer rows.Close()`, turn the skip indexes collection into an anonymous function you call directly, move `existingIndexes` definition outside this function.

vincentbernat (changes requested):
Almost there. Also add an entry to the changelog.

orchestrator/clickhouse/migrations_helpers.go
PolyAnthi: https://github.com/akvorado/akvorado/pull/2368/commits/1ffd17492831fd405780bd3b53aa1ad5ab5d9c71

vincentbernat (approved):

vincentbernat:
Thanks! It should be quite useful.

PR #2381, opened 2026-04-20
build: ship akvorado-inlet

vincentbernat:
It will be more interesting on DCE work for real. See #2112.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /bin/akvorado
+/bin/akvorado-inlet
 /test/
 /orchestrator/clickhouse/data/asns.csv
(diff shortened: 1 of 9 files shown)


vincentbernat:
It does not work. In addition to a new version of github.com/go-playground/validator/10, we also need a new version of google.golang.org/protobuf (see https://github.com/golang/protobuf/issues/1704). With both, we get the inlet at 35M, which is still big. One of the culprit is github.com/gin-gonic/gin, which pulls github.com/ugorji/go/codec.

vincentbernat:
After removing Gin, we are able to reach 26M. I don't see any obvious additional gain.

image

PR #2382, opened 2026-04-22
feat: Support SrcCommunities and SrcLargeCommunities columns with filtering in Akvorado

fabianbees:
Currently Akvorado only exposes destination‑community information (`DstCommunities` / `DstLargeCommunities`). Since the destination columns already exist, it makes sense to add equivalent source‑community columns.

Adding `SrcCommunities` and `SrcLargeCommunities` would be a huge benefit for our use case. With these columns we could:

- Distinguish different kinds of traffic (e.g., traffic from different products we offer).
- View and filter custom *source* BGP community data, such as customer‑specific tags or product-type identifiers.

The implementation should be straightforward and introduce minimal overhead: the required information is already available in the enricher, and we only need to store it in ClickHouse. I have tried to keep the code similar to the existing `DstCommunities` implementation, but please review it to ensure I haven’t missed anything critical.

I’m not very familiar with your testing suite/setup, so feel free to add any tests you think are necessary, or let me know what steps are needed to get this feature merged.


vincentbernat:
It was not included because the information could be incorrect since the router uses only routes to send traffic, not receive (and communities are attached to routes).

vincentbernat:
I have been thinking a bit and I think an alternative could be:
1. define these columns but make them disabled by default
2. document them specifically in `02-configuration.md`, so that people know how to enable with a risk that the information they contain may be incorrect

I have updated a bit the code in schema, so that it's easier to add them. Just change DstCommunities and DstLargeCommunities definitions to SrcCommunities and SrcLargeCommunities and they are already disabled, so it should be fine.

fabianbees:
This will work for us, I think a feature flag could be a very good option.

> It was not included because the information could be incorrect since the router uses only routes to send traffic, not receive.

That’s a valid concern. However, the same limitation also applies to the `SrcAS` number — it is derived from routing information as well, so it could be incorrect for the same reason. I agree that any potentially inaccurate data should be clearly documented.

For our use‑case, a `SrcCommunities` field remains the best option for tagging additional information to IP addresses, currently I can’t think of a better option than tagging with BGP communities, which can then be filtered in Akvorado.

common/schema/definition.go
```diff
@@ -431,6 +433,18 @@ END`,
...
+				ParserType:         "community",
+				ClickHouseMainOnly: true,
+				ClickHouseType:     "Array(UInt32)",
+			},
+			{
+				Key:                 ColumnSrcLargeCommunities,
+				ClickHouseMainOnly:  true,
+				ClickHouseType:      "Array(UInt128)",
+				ConsoleNotDimension: true,
+			},
```
vincentbernat, line 408: You can remove ColumnDstCommunities and ColumnDstLargeCommunities. They will automatically be derived from these ones.

common/schema/definition_gen.sh
```diff
@@ -24,7 +24,7 @@ $(sed -En '/.*ColumnKey = iota/,/^[[:space:]]ColumnLast/p' common/schema/definit
     | sed \$d \
     | sed -En 's/^[[:space:]]+(Column[^ ]+).*/\1/p' \
     | sed -E \
-             -e 's/(ColumnDst(ASPath|Communities|LargeCommunities|[123]..AS))/\1: 0,/;t' \
+             -e 's/(ColumnSrc(Communities|LargeCommunities)|ColumnDst(ASPath|Communities|LargeCommunities|[123]..AS))/\1: 0,/;t' \
```
vincentbernat: Instead, you need to remove Communities and LargeCommunities from this regex.

console/query/column.go
```diff
@@ -120,8 +120,20 @@ func (qc Column) ToSQLSelect(sch *schema.Component) string {
...
+								toString(bitAnd(c, 0xffffffff))), SrcLargeCommunities)
+			), ' ')`
 	case schema.ColumnDstCommunities:
-		strValue = `arrayStringConcat(arrayConcat(arrayMap(c -> concat(toString(bitShiftRight(c, 16)), ':', toString(bitAnd(c, 0xffff))), DstCommunities), arrayMap(c -> concat(toString(bitAnd(bitShiftRight(c, 64), 0xffffffff)), ':', toString(bitAnd(bitShiftRight(c, 32), 0xffffffff)), ':', toString(bitAnd(c, 0xffffffff))), DstLargeCommunities)), ' ')`
+		strValue = `arrayStringConcat(arrayConcat(
+			arrayMap(c -> concat(toString(bitShiftRight(c, 16)), ':', toString(bitAnd(c, 0xffff))), DstCommunities),
+			arrayMap(c -> concat(toString(bitAnd(bitShiftRight(c, 64), 0xffffffff)), ':',
+								toString(bitAnd(bitShiftRight(c, 32), 0xffffffff)), ':',
+								toString(bitAnd(c, 0xffffffff))), DstLargeCommunities)
+			), ' ')`
```
vincentbernat: This could be `case schema.ColumnSrcCommunities, schema.ColumnDstCommunities` and use `qc.String()[:3]` instead of `Dst`.

console/filter.go
```diff
@@ -206,6 +206,51 @@ LIMIT %d`, columnName, columnName, input.Limit)
...
+				break
+			}
+			for _, result := range results {
+				completions = append(completions, filterCompletion{
+					Label:  result.Label,
+					Detail: result.Detail,
+					Quoted: false,
+				})
+			}
+			input.Prefix = ""
```
vincentbernat: Same here. We want to use the same code for `srccommunities` and `dstcommunities`.

outlet/core/enricher.go
```diff
@@ -141,8 +141,19 @@ func (w *worker) enrichFlow(exporterIP netip.Addr, exporterStr string) bool {
...
+	if len(sourceRouting.LargeCommunities) > 0 {
+		communitiesSrc := make([]schema.UInt128, len(sourceRouting.LargeCommunities))
+		for i, comm := range sourceRouting.LargeCommunities {
+			communitiesSrc[i] = schema.UInt128{
+				High: uint64(comm.ASN),
+				Low:  (uint64(comm.LocalData1) << 32) + uint64(comm.LocalData2),
+			}
+		}
+		flow.AppendArrayUInt128(schema.ColumnSrcLargeCommunities, communitiesSrc)
+	}
```
vincentbernat, line 151: Same here. We need a function to avoid repeating the same code twice.

console/widgets.go
```diff
@@ -22,11 +22,10 @@ func (c *Component) widgetFlowLastHandlerFunc(gc *gin.Context) {
...
 	}{
-		{schema.ColumnDstCommunities, `arrayMap(c -> concat(toString(bitShiftRight(c, 16)), ':',
-                      toString(bitAnd(c, 0xffff))), DstCommunities)`},
-		{schema.ColumnDstLargeCommunities, `arrayMap(c -> concat(toString(bitAnd(bitShiftRight(c, 64), 0xffffffff)), ':',
-                      toString(bitAnd(bitShiftRight(c, 32), 0xffffffff)), ':',
-                      toString(bitAnd(c, 0xffffffff))), DstLargeCommunities)`},
+		{schema.ColumnSrcCommunities, `arrayMap(c -> concat(toString(bitShiftRight(c, 16)), ':', toString(bitAnd(c, 0xffff))), SrcCommunities)`},
+		{schema.ColumnSrcLargeCommunities, `arrayMap(c -> concat(toString(bitAnd(bitShiftRight(c, 64), 0xffffffff)), ':', toString(bitAnd(bitShiftRight(c, 32), 0xffffffff)), ':', toString(bitAnd(c, 0xffffffff))), SrcLargeCommunities)`},
+		{schema.ColumnDstCommunities, `arrayMap(c -> concat(toString(bitShiftRight(c, 16)), ':', toString(bitAnd(c, 0xffff))), DstCommunities)`},
+		{schema.ColumnDstLargeCommunities, `arrayMap(c -> concat(toString(bitAnd(bitShiftRight(c, 64), 0xffffffff)), ':', toString(bitAnd(bitShiftRight(c, 32), 0xffffffff)), ':', toString(bitAnd(c, 0xffffffff))), DstLargeCommunities)`},
```
vincentbernat: Abstract both `arrayMap(...)` with local functions `communitiesFormat(prefix string)` and `largeCommunitiesFormat(prefix string)`.

vincentbernat (changes requested):
Also, please, rebase on main.

console/filter.go
```diff
@@ -205,18 +205,19 @@ LIMIT %d`, columnName, columnName, input.Limit)
...
 			}{}
+			columnNamePrefix := c.fixQueryColumnName(input.Column)[:3]
 			sqlQuery := fmt.Sprintf(`
 SELECT label, detail FROM (
  SELECT
   'community' AS detail,
   concat(toString(bitShiftRight(c, 16)), ':', toString(bitAnd(c, 0xffff))) AS label
  FROM (
-  SELECT arrayJoin(SrcCommunities) AS c
+  SELECT arrayJoin(concat(toString(%s), 'Communities')) AS c
```
vincentbernat: You can use `arrayJoin(%sCommunities)`

console/filter.go
```diff
@@ -229,60 +230,15 @@ SELECT label, detail FROM (
   'large community' AS detail,
   concat(toString(bitAnd(bitShiftRight(c, 64), 0xffffffff)), ':', toString(bitAnd(bitShiftRight(c, 32), 0xffffffff)), ':', toString(bitAnd(c, 0xffffffff))) AS label
  FROM (
-  SELECT arrayJoin(SrcLargeCommunities) AS c
+  SELECT arrayJoin(concat(toString(%s), 'LargeCommunities')) AS c
```
vincentbernat: Same here.

vincentbernat (changes requested):
Almost done!

fabianbees:
> Almost done!

Okay, I think I’ve addressed all of your comments so far. And thank you for your review, by the way.

vincentbernat:
OK, now, you have the failing tests. Some of them are because you did change some SQL strings by adding more whitespace. I can also fix that myself if you prefer.

vincentbernat:
I have pushed one additional commit to disable `SrcCommunities` and `SrcLargeCommunities` by default.

PR #2433, opened 2026-05-22
More efficient RIB tree for BMP

vincentbernat:
While writing a blog post about RIB sharding, I find myself looking a bit more at the code and was thinking the global lock looks very bad. Since we now have a concurrent benchmark, we can look again at using "persist" variants from bart package. It works well! -60% read, -30% write! There was a trick not used in the previous tentatives: not using ModifyPersist which triggers "false" modifications, but instead use `Get()` + `InsertPersist()` or `DeletePersist()`.

The second change is unrelated as it fixes a bug that was already there: it was possible to fetch a stale index from the tree and retrieve the wrong routes if the index is recycled. The third commit of this PR fixes that by introducing a generation number to detect when a prefix is recycled. It does not seem to impact performance.

cc @bogi788 if you want to test (but I need to spend a bit more time to ensure everything is correct, so no hurry).

### DIFF
--- a/common/helpers/intern/intern.go
+++ b/common/helpers/intern/intern.go
@@ -51,8 +51,8 @@ type Pool[T Value[T]] struct {
 // keeping to the raw value.
 type internValue[T Value[T]] struct {
 	next     Reference[T] // next value with the same hash
-	previous Reference[T] // previous value with the same hash
 	refCount uint32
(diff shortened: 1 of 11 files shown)


vincentbernat:
comparison

vincentbernat:
Explanations in this [blog post](https://vincent.bernat.ch/en/blog/2026-akvorado-rib-sharding).

PR #2435, opened 2026-05-24
outlet/clickhouse: add a hook to allow rebalancing on retries

vincentbernat:
When ClickHouse is unavailable for some reason, we shouldn't block rebalancing.

But I don't think this is a good idea: if we cannot join ClickHouse, there is no point in letting rebalancing happens, we will just loose more data.

Cc: @bleuchtang
Related: #1908

I have also pushed a few commits in main (also included in this PR), but I don't think it will help your case.


vincentbernat:
I'll close this one as I don't think this is a good idea.

PR #2458, opened 2026-05-30
docker: add option to use GeoLite2-City

edeka-spatt:
Add option to use GeoLite2-City in docker-compose-local.yml

https://github.com/akvorado/akvorado/discussions/2453


vincentbernat:
Thanks!

PR #2462, opened 2026-06-02
outlet/dns: add reverse DNS enrichment

VitisEK:
﻿## Summary

This adds optional reverse DNS enrichment in the outlet pipeline. When enabled, the outlet resolves PTR records for `SrcAddr` and `DstAddr` asynchronously and stores cache hits in new `SrcDNSName` and `DstDNSName` dimensions.

The feature is disabled by default and does not change inlet behavior. Cache misses never block flow processing: the current flow keeps empty DNS name fields while the IP is queued for asynchronous resolution.

## Implementation

- add a new `outlet/dns` component with explicit resolvers, timeout/attempts, positive and negative caches, bounded cache size, async queueing, pending query deduplication, and concurrent query limits
- use `github.com/miekg/dns` for PTR lookups so positive cache TTLs can be derived from DNS answers and clamped by config
- integrate the component into outlet startup and `outlet/core/enricher.go`
- add `SrcDNSName` and `DstDNSName` as normal string dimensions
- document the new `outlet.dns` configuration and default-disabled behavior

## Notes

This intentionally does not use the metadata provider and does not perform ClickHouse mutations for historical flow rows. DNS enrichment only applies to newly processed flows.

## Tests

- `go test -count=1 ./outlet/dns`
- Tested in a normal build environment: the project builds successfully, and the DNS enrichment path works as expected.

Additional checks attempted in the Codex Windows checkout were blocked by missing generated artifacts:

- `go test ./common/schema` fails because `common/clickhousedb/mocks/mock_driver.go` is missing
- `go test ./outlet/core` fails because protobuf generated types and ClickHouse mocks are missing; on Windows it also hits an existing BMP provider `syscall.Handle` compile issue
- `go test ./cmd/akvorado` fails because `common/embed/data/embed.zip` is missing


vincentbernat:
It looks OK and useful. About not blocking, I don't know if it wouldn't be better to block as maybe it's better to be slow at start but not having partial information. Or maybe make it configurable? Or maybe two timeouts? One blocking, one non-blocking?

VitisEK:
@vincentbernat Thanks for the feedback! I updated the PR to make this configurable while keeping the current non-blocking behavior as the default.

The DNS configuration now has two separate timeout concepts:

- `timeout`: still the timeout for an actual DNS query attempt
- `initial-timeout`: the maximum time the outlet hot path may wait on a cache miss, only when enabled

By default, `wait-for-initial-result` is `false`, so cache misses keep the original asynchronous behavior. When set to `true`, concurrent lookups for the same IP share the same pending PTR query and wait up to `initial-timeout`; if the result arrives in time, the first flow can be enriched, otherwise processing continues without blocking further and the async lookup still populates the cache.

I also added tests for the default non-blocking behavior, successful initial wait, initial timeout followed by later cache hit, pending-query deduplication with concurrent callers, negative cache behavior, and queue-full cleanup.

vincentbernat:
You can do everything with two timeouts:
- `async-timeout` is the total timeout
- `sync-timeout` is the timeout of the sync part

If you want the async behavior only, set `sync-timeout` to 0.

Sven286:
I'm looking forward to use this feature. I think it's very usefull.
Can anyone estimate when this PR is merged?

PR #2465, opened 2026-06-04
outlet/flow: add PBB decapsulation

VitisEK:
## Summary

This adds `pbb` as a decapsulation protocol for sampled packet decoding. It supports IEEE 802.1ah Provider Backbone Bridges / MAC-in-MAC, including SPB-M deployments.

When enabled, Akvorado decapsulates the outer PBB frame, extracts the 24-bit I-SID into a new `PBBISID` L2 column, and continues decoding the inner Ethernet frame.

## Implementation

- add `DECAP_PBB` and expose it as `decapsulation-protocol: pbb`
- accept `spbm` and `mac-in-mac` as configuration aliases for `pbb`
- add EtherTypes for 802.1ad and PBB
- parse PBB I-TAG and extract I-SID
- decode the inner Ethernet frame after PBB decapsulation
- add disabled-by-default L2 dimension `PBBISID`
- document `pbb` in the decapsulation protocol list

If `pbb` is configured but the sampled packet is a normal Ethernet frame, decoding falls back to native Ethernet parsing. This keeps mixed captures usable.

## Tests

- `git diff --check`
- `go test ./common/pb`
- `go test ./common/schema`
- `go test ./outlet/flow/decoder`
- `go test ./inlet/flow`
- `go test ./outlet/flow/decoder/sflow`
- `go test ./outlet/flow/decoder/netflow`

### DIFF
--- a/common/constants/ether.go
+++ b/common/constants/ether.go
@@ -12,4 +12,8 @@ const (
 	ETypeMPLS = 0x8847
 	// ETypeVLAN is the ether type for Dot1Q
 	ETypeVLAN = 0x8100
+	// ETypeDot1AD is the ether type for Dot1ad
+	ETypeDot1AD = 0x88a8
+	// ETypePBB is the ether type for Provider Backbone Bridges
+	ETypePBB = 0x88e7
 )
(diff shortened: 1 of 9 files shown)


vincentbernat:
For everything else, mixed captures are not allowed. This is by design because otherwise, we may decap some "user" packets, so it ensures the user sends the right encapsulation to the right endpoint. When they expect one specific encap, they send it to a specific port and another specific encap to another port.

Also, for testing, I prefer to have PCAP files. It's easier to inspect in Wireshark.

vincentbernat:
Also, maybe it shouldn't be a decap behavior at all. For GRE and others, we want to be able to make a difference between a wanted GRE tunnel or an unwanted/customer GRE tunnel. For MAC, it seems less useful. You could just remove the encap and store the data if the right columns are enabled.

Issue #2464, opened 2026-06-03
Vlan-based interface classification not working with sFlow for outbound flows


plumbeo:
### Disclaimer

- [x] I have read and understood the disclaimer.
- [x] I am a new user.

### Akvorado version

2.4.0

### Bug description

I've been testing akvorado with a device that supports sFlow but doesn't generate extended switch data and I'm having a problem when I try to classify the interfaces according to the vlans.

I'm sampling an interface in both directions (egress and ingress) and I'm using these classifier rules: 

```
    - |
      Exporter.IP == "192.0.2.1" && Interface.Name == "GigabitEthernet0/0/1" && Interface.VLAN == 100 &&
      ClassifyConnectivity("ix") &&
      ClassifyProvider("provider100") &&
      ClassifyExternal()
    - |
      Exporter.IP == "192.0.2.1" && Interface.Name == "GigabitEthernet0/0/1" && Interface.VLAN == 200 &&
      ClassifyConnectivity("ix") &&
      ClassifyProvider("provider200") &&
      ClassifyExternal()
    - |
```
The interfaces are classified as expected for inbound flows but not for outbound flows, so they end up as internal interfaces and are excluded by the default filter and even if I add a fallback classifier to at least set the correct boundary the outbound flows don't match the inbound ones and the traffic shows up as "other" in every bidirectional graph.

Looking to the flows it looks like the only available info for the outbound flows on these interfaces are ``OutIfName`` and ``OutIfSpeed`` and only ``SrcVlan`` shows the vlan id while ``DstVlan`` is empty.

plumbeo:
[...]

PR #2466, opened 2026-06-05
Populate DstVlan from the vlan id of the ethernet frame

plumbeo:
If we don't receive more precise information from the protocol (for example when using sFlow without extended switch data) use the decoded vlan id of the ethernet frame to populate the ``DstVlan`` field of the flow, making sure that we can match on ``Interface.VLAN`` to classify an interface even for outbound flows when doing egress sampling.

Without this change if we sample on ingress and egress on the same interface any bidirectional graph shows the outbound traffic as "other" because the wrongly classified or unclassified outbound interface doesn't match any inbound interface.

Discussed in https://github.com/akvorado/akvorado/issues/2464.


vincentbernat:
Thanks!

Issue #54, opened 2022-07-26
VRF support


vincentbernat:
Add VRF support in Akvorado. I don't have use for this myself but it may be relevant in the future. For Netflow, the VRF is provided as part of the flow and the name of the VRF is provided as an option. Therefore, it is straightforward to add. The mapping between networks and names/roles/sites/regions/tenants also need to be updated to use VRFs. If #52 is implemented, it would also need to be VRF-aware.

barhom:
We are encountering this issue now.
We have a customer that we send data into a VRF for and then on our edge routers traffic that is within this VRF is not counted because it is lacking AS information.

Do you know of a good solution?

dimage1:
Is it possible to propagate new IPFIX - Type: ingressVRFID (234), Length: 4 - into the flow fields ? Absence of it reduces our observability options on a datacenter flows.

dimage1:
Hi @vincentbernat ! Opened #2488 with the lightweight VRF support discussed above.

It decodes the standard IPFIX fields `ingressVRFID` (234) and `egressVRFID` (235) into disabled-by-default flow columns: `IngressVRFID` and `EgressVRFID`.

This does not try to solve the broader VRF-aware network classification or routing enrichment parts yet. The goal is to expose the raw VRF IDs in Akvorado first, so deployments with stable VRF IDs can filter/group flows and add their own enrichment on top.

PR #2488, opened 2026-06-23
outlet/flow: decode IPFIX VRF IDs

dimage1:
Add support for decoding IPFIX VRF IDs into optional flow columns:

- ingressVRFID (234) -> IngressVRFID
- egressVRFID (235) -> EgressVRFID

The new columns are disabled by default and can be enabled through the schema
configuration, like other optional flow fields.

This covers the raw VRF ID fields from flow records. It does not implement
VRF-aware network classification or routing enrichment.

Refs #54

Tests:
- go test ./common/schema ./outlet/flow/decoder/netflow
- go test ./console/query

### DIFF
--- a/common/schema/definition.go
+++ b/common/schema/definition.go
@@ -220,6 +220,8 @@ const (
 	ColumnMPLS2ndLabel
 	ColumnMPLS3rdLabel
 	ColumnMPLS4thLabel
+	ColumnIngressVRFID
+	ColumnEgressVRFID
 
 	// ColumnLast points to after the last static column, custom dictionaries
 	// (dynamic columns) come after ColumnLast
@@ -569,6 +571,8 @@ END`,
 				ClickHouseAlias:    "MPLSLabels[4]",
 				ParserType:         "uint",
 			},
+			{Key: ColumnIngressVRFID, Disabled: true, ParserType: "uint", ClickHouseType: "UInt32"},
+			{Key: ColumnEgressVRFID, Disabled: true, ParserType: "uint", ClickHouseType: "UInt32"},
 		},
 	}.finalize()
 }
--- a/common/schema/testdata/schema.yaml
+++ b/common/schema/testdata/schema.yaml
@@ -503,3 +503,11 @@
   clickhousetype: UInt32
   clickhousealias: MPLSLabels[4]
   clickhousemainonly: true
+- key: IngressVRFID
+  name: IngressVRFID
+  parsertype: uint
+  clickhousetype: UInt32
+- key: EgressVRFID
+  name: EgressVRFID
+  parsertype: uint
+  clickhousetype: UInt32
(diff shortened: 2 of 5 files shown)


outlet/flow/decoder/netflow/decode_test.go
```diff
@@ -52,3 +57,31 @@ func TestDecodeUNumber(t *testing.T) {
...
+		t.Fatalf("decodeRecord() decoded %d flows, expected 1", len(*got))
+	}
+	flow := (*got)[0]
+	if got := flow.OtherColumns[schema.ColumnIngressVRFID]; got != uint32(42) {
+		t.Fatalf("IngressVRFID = %v, expected %v", got, uint32(42))
+	}
+	if got := flow.OtherColumns[schema.ColumnEgressVRFID]; got != uint32(43) {
+		t.Fatalf("EgressVRFID = %v, expected %v", got, uint32(43))
+	}
+}
```
vincentbernat: Tests should be against real packets (in the PCAP format). The test looks a bit artificial otherwise. This is already covered in the tests you modified below.

vincentbernat:
Are these IDs stable? If you unconfigure/reconfigure IPFIX or restart the router, do they stay the same?

dimage1:
For our deployment, they are stable for the lifetime of a virtual network. The
VRF ID is allocated by our control plane when the network is created and then
propagated to the routers/dataplane. It is not derived from the IPFIX exporter
state, so reconfiguring IPFIX or restarting a router does not change it; the
router is reconfigured with the same control-plane-provided VRF/table ID.

I would not claim these IDs are globally meaningful forever. They are
exporter/control-plane local IDs, and in our case they may be released when the
network is deleted. This PR only exposes the raw IPFIX value as an optional
column, so deployments where the ID is meaningful can filter/group/enrich on it.

This also matches the IANA IPFIX registry semantics: `ingressVRFID`/`egressVRFID`
are `unsigned32` identifiers, unique per Metering Process.

outlet/flow/decoder/netflow/decode_test.go
dimage1: make sense, re-pushed with small changes

vincentbernat:
Thanks!

dimage1:
@vincentbernat hi ! Is any release expected soon, so we can start using the feature in the official Akvorado version? The last release was in may, I expected them be around once per month. Thanks !

vincentbernat:
You just need to use `:latest` (or pin to a specific hash) in `docker/versions.yml` to use a newer version. You don't need to wait for an official release.

dimage1:
@vincentbernat :latest is still on the 2.4 from May. And all the commits after this shows just Skipped in the publishing state : CI / 🚀 Publish release (push) Skipped . It might be that there's no released suitable version with VRF change yet.

vincentbernat:
Sorry, I meant `:main` or a specific hash like `quay.io/akvorado/akvorado@sha256:acc384b46e3e3f9d740f2bac585c9b8a287e4399ba10ca464abcb3ade2ac4e51`

PR #2489, opened 2026-06-28
Add support for EoMPLS with or without Control Word

nitzan-tz:
This PR add support for EoMPLS (In my case EVPN over MPLS) 
My test file is for case where there is a control word between the last label and the next ethernet but there is also support for the other option without it


outlet/flow/decoder/netflow/decode_evpn_test.go
```diff
@@ -0,0 +1,110 @@
...
+	})
+
+	reader, err := pcapgo.NewReader(f)
+	if err != nil {
+		t.Fatalf("NewReader(%q) error:\n%+v", pcapfile, err)
+	}
+	return gopacket.NewPacketSource(reader, layers.LayerTypeEthernet)
+}
+
+func TestDecodeEVPN(t *testing.T) {
```
vincentbernat: Move it back with the others. Don't declare a readPcapLocal(), resue the existing one. You need to have one packet per PCAP files.

outlet/flow/decoder/helpers.go
```diff
@@ -234,11 +234,26 @@ func ParseEthernet(sch *schema.Component, bf *schema.FlowMessage, decap pb.RawFl
...
 					etherType = constants.ETypeIPv4
 				case 6:
 					etherType = constants.ETypeIPv6
 				default:
+					if len(data) >= 14 {
+						if len(mplsLabels) > 0 && decap == pb.RawFlow_DECAP_NONE {
+							bf.AppendArrayUInt32(schema.ColumnMPLSLabels, mplsLabels)
+						}
+						return ParseEthernet(sch, bf, decap, data)
+					}
```
vincentbernat, line 271: It seems a bit odd to make everything else Ethernet. I'll need to check a bit of documentation to understand how it works.

vincentbernat (changes requested):

vincentbernat:
I think this is worse than before. My proposition:
- require bottom == 1
- label 0 means IPv4 payload, label 2 means IPv6 payload
- for another label, sniff first nibble: 4 ⇒ IPv4 payload, 6 ⇒ IPv6 payload, 0 ⇒ CW
- if CW is 0, then Ethernet
- otherwise give up

As a sidenote, I see even Wireshark is confused between flow 3 (correctly parsed) and flow 4 (not correctly parsed, while it seems to contain also a CW).

nitzan-tz:
I did the changes to parse 
Verify that the last lable is the BoS
Check explicit null for IPv4 and IPv6 
After that check for 0x4, 0x6 or CW and if CW parse ethernet 

I also add PCAPs for the test cases 
BTW for me wireshark parse both flow 3 and flow 4

vincentbernat:
I have merged your changes in 61e6cf7ae06b and added 30a9e85e6715 on top of it. I am closing the PR as GitHub does not detect that.

Thanks!

Issue #2483, opened 2026-06-15
outlet: optional Kafka output for enriched flows?


neumayer:
### Disclaimer

- [x] I have read and understood the disclaimer.
- [x] I plan to implement the feature myself.

### Akvorado version

2.0

### Feature description

Would you be open to an optional Kafka output on the outlet, that sends the decoded and enriched flow to a Kafka topic in addition to ClickHouse? Off by default, so nothing changes for existing setups.

What we're after is just getting the enriched flows out, so other systems can consume them more or less in real time. Right now neither option is great: reading the inlet topic means we'd have to re-do the NetFlow/IPFIX decoding ourselves, and a materialized view / export from ClickHouse goes against what ClickHouse is good at and adds load on it. The outlet already has the decoded and enriched flow in hand, so it seems like the right place to branch off. The actual analysis stays downstream, outside Akvorado — so this is really just plumbing, not detection. I think it also fits the direction in #2006, and the "do it downstream, against the data" approach from #53 / your [DDoS + flowspec post](https://vincent.bernat.ch/en/blog/2023-akvorado-ddos-flowspec).

A few questions before I build anything real:

1. Is this something you'd consider taking upstream at all?
2. What shape would you prefer — a dedicated Kafka output, or making the sink pluggable (ClickHouse + Kafka) so more outputs can be added later? Your call.
3. For payload and delivery: flat JSON or a versioned protobuf like the inlet topic, and best-effort or at-least-once?

Happy to implement it however you'd want. I did a small feasibility spike and it looks like a fairly contained change — I'd keep it opt-in and minimal so it doesn't add much to maintain.

vincentbernat:
The ClickHouse output is heavily engraved into the outlet for performance reason. If we try to do Kafka+ClickHouse, it may lead either to loss of performance or complexity. We could restore the code to write Protobuf efficiently that was there before Akvorado 2.0 to keep something efficient. The code would be active only if there is a Kafka output. We could even have something modular. The encoding of the flows would depend on the active outputs. But this is too complex and I still have to refactor some stuff.

I would suggest to stream to Kafka to ClickHouse to sidestep all these issues, but you have already considered it and are not OK with it.

Note that I also plan to make flow encoding a bit more modular, so it may be possible to do something "simpler" in the future (there will be some complexity, but it would be elsewhere). But I already have a ton of stuff to implement before.

neumayer:
[...]

PR #2520, opened 2026-07-06
outlet: optional Kafka output for enriched flows

neumayer:
Adds an optional second output to the outlet: the decoded and enriched flow is produced to a Kafka topic in parallel with the ClickHouse insert, so downstream systems can consume enriched flows in near real time without re-decoding the inlet topic or exporting from ClickHouse. Disabled by default; no-op when off.

Follows the discussion in #2483.

## Approach

Restores the pre-2.0 append-time Protobuf encoding, gated behind `kafka-output.enabled`. When enabled, each `Append*` also writes the field into a per-flow Protobuf buffer, with field numbers and types derived in `finalize()`. Full fidelity, including arrays (AS path, communities). A JSON read-back variant was tried first and dropped: this is ~4.7× faster and allocates ~50× less.

Behaviour worth calling out:

- **Delivery is best-effort and at-most-once.** `Send()` hands the record to `TryProduce()`, so it never blocks; when the producer buffer is full the record is dropped and counted. The ClickHouse path can therefore never be backpressured by this output — which is the property that made it acceptable to put on the hot path at all.
- **The topic carries the schema hash** (`topic-<hash>`), the same hash as in the generated `FlowMessagev<hash>` message name, so an incompatible schema change lands on a new topic instead of mixing wire layouts for consumers. Mirrors the inlet topic's version suffix. There is deliberately no opt-out.
- **The orchestrator manages the output topic** when a top-level `kafka-output` block is set — a peer of `kafka` with its own connection, so the output topic can live on a different cluster. Without it, the outlet relies on broker auto-creation.
- **On shutdown** the buffered records are flushed within `shutdown-timeout` (default `1s`) rather than discarded.
- **The `.proto` is served** at `/api/v0/outlet/kafka-output/schema.proto`, so consumers can fetch the definition matching the running schema.

```yaml
kafka-output:
  enabled: true
  topic: flows-enriched
  brokers:
    - kafka:9092
  queue-size: 4096
```

## Measurements

Validated in staging against a production flow stream peaking ~150k flows/s — 7 days on the original producer, plus a shorter run since the `TryProduce()` rework (a 24 h read across a full daily peak is still outstanding).

- **No ClickHouse degradation.** Insert time, wait time, consumer lag and outlet CPU/memory stay flat with the output enabled, confirmed under overload. This is the property drop-on-full exists to guarantee, and it holds on both producers.
- **Cost:** ~+1 allocation per flow, no measurable CPU delta, 113 B wire payload (265 B for the JSON variant that was tried first).
- **Latency:** p50 ~0.7 s, p99 ~2 s end-to-end, load-independent from 81k–130k flows/s, no decode errors.
- **Drops:** none observed at `queue-size` 65536 and above, up to 150,016 msg/s; 16Ki dropped 52–64% at peak. An earlier, shorter capture saw a ~0.1% residual at 65536, so read this as "not observed" rather than proven absent.
- **Sizing `queue-size`:** the producer's steady-state working set is only ~7.7k records, and unchanged across the rework. The 16Ki failure was burst absorption, not steady state — so the 4096 default is in the right territory, and the knob really only buys burst and broker-stall tolerance.

Producer *linger* did not help (during bursts the client already fills batches to the size cap and bypasses it) and was removed. A Protobuf-decoding consumer tracked the producer with no gaps, which exercises the schema-hash topic naming and the field numbering end to end.

## Metrics

Prefixed `akvorado_outlet_kafkaoutput_`: `sent_messages_total`, `sent_bytes_total`, `dropped_messages_total`, `errors_total`, plus the franz-go client metrics via kprom — notably `buffered_produce_records_total` (producer buffer depth), `request_duration_e2e_seconds` and `request_throttled_seconds`.

`dropped_messages_total` is the reliable saturation signal; the buffer gauge is a scrape-time snapshot and can miss brief bursts.

## Note on the commits

This branch includes @vincentbernat's own commits from review — the component renames, configurable compression and load balancing, `TryProduce()`, the HTTP schema endpoint, the orchestrator topic fix and the gcassert replacement. Happy to consolidate some of them before merge if you'd prefer a tidier history.

The changelog entries are added, including a `💥` for the metric rename: the `outlet/kafka` → `outlet/kafkainput` package move renames the ten metrics that package exports, since the prefix derives from the package path. The configuration key keeps working through a rename hook, but metric names have no compatibility path, so anyone with their own dashboards or alerts on `akvorado_outlet_kafka_*` needs to update them.

### DIFF
--- a/cmd/akvorado/outlet.go
+++ b/cmd/akvorado/outlet.go
@@ -20,6 +20,7 @@ import (
 	"akvorado/outlet/core"
 	"akvorado/outlet/flow"
 	"akvorado/outlet/kafka"
+	"akvorado/outlet/kafkaout"
 	"akvorado/outlet/metadata"
 	"akvorado/outlet/metadata/provider/snmp"
 	"akvorado/outlet/routing"
@@ -35,6 +36,7 @@ type OutletConfiguration struct {
 	Kafka        kafka.Configuration
 	ClickHouseDB clickhousedb.Configuration
 	ClickHouse   clickhouse.Configuration
+	KafkaOut     kafkaout.Configuration
 	Flow         flow.Configuration
 	Core         core.Configuration
 	Schema       schema.Configuration
@@ -50,6 +52,7 @@ func (c *OutletConfiguration) Reset() {
 		Kafka:        kafka.DefaultConfiguration(),
 		ClickHouseDB: clickhousedb.DefaultConfiguration(),
 		ClickHouse:   clickhouse.DefaultConfiguration(),
+		KafkaOut:     kafkaout.DefaultConfiguration(),
 		Flow:         flow.DefaultConfiguration(),
 		Core:         core.DefaultConfiguration(),
 		Schema:       schema.DefaultConfiguration(),
@@ -149,6 +152,13 @@ func outletStart(r *reporter.Reporter, config OutletConfiguration, checkOnly boo
 	if err != nil {
 		return fmt.Errorf("unable to initialize outlet ClickHouse component: %w", err)
 	}
+	kafkaOutComponent, err := kafkaout.New(r, config.KafkaOut, kafkaout.Dependencies{
+		Daemon: daemonComponent,
+		Schema: schemaComponent,
+	})
+	if err != nil {
+		return fmt.Errorf("unable to initialize Kafka output component: %w", err)
+	}
 	coreComponent, err := core.New(r, config.Core, core.Dependencies{
 		Daemon:     daemonComponent,
 		Flow:       flowComponent,
@@ -156,6 +166,7 @@ func outletStart(r *reporter.Reporter, config OutletConfiguration, checkOnly boo
 		Routing:    routingComponent,
 		Kafka:      kafkaComponent,
 		ClickHouse: clickhouseComponent,
+		KafkaOut:   kafkaOutComponent,
 		HTTP:       httpComponent,
 		Schema:     schemaComponent,
 	})
@@ -181,6 +192,7 @@ func outletStart(r *reporter.Reporter, config OutletConfiguration, checkOnly boo
 		metadataComponent,
 		routingComponent,
 		kafkaComponent,
+		kafkaOutComponent,
 		coreComponent,
 	}
 	return StartStopComponents(r, daemonComponent, components)
--- a/cmd/akvorado/outlet.go
+++ b/cmd/akvorado/outlet.go
@@ -154,6 +154,7 @@ func outletStart(r *reporter.Reporter, config OutletConfiguration, checkOnly boo
 	}
 	kafkaOutputComponent, err := kafkaoutput.New(r, config.KafkaOutput, kafkaoutput.Dependencies{
 		Daemon: daemonComponent,
+		HTTP:   httpComponent,
 		Schema: schemaComponent,
 	})
 	if err != nil {
--- a/cmd/akvorado/outlet.go
+++ b/cmd/akvorado/outlet.go
@@ -19,8 +19,8 @@ import (
 	"akvorado/outlet/clickhouse"
 	"akvorado/outlet/core"
 	"akvorado/outlet/flow"
-	"akvorado/outlet/kafka"
-	"akvorado/outlet/kafkaout"
+	"akvorado/outlet/kafkainput"
+	"akvorado/outlet/kafkaoutput"
 	"akvorado/outlet/metadata"
 	"akvorado/outlet/metadata/provider/snmp"
 	"akvorado/outlet/routing"
@@ -33,10 +33,10 @@ type OutletConfiguration struct {
 	HTTP         httpserver.Configuration
 	Metadata     metadata.Configuration
 	Routing      routing.Configuration
-	Kafka        kafka.Configuration
+	Kafka        kafkainput.Configuration
+	KafkaOutput  kafkaoutput.Configuration
 	ClickHouseDB clickhousedb.Configuration
 	ClickHouse   clickhouse.Configuration
-	KafkaOut     kafkaout.Configuration
 	Flow         flow.Configuration
 	Core         core.Configuration
 	Schema       schema.Configuration
@@ -49,10 +49,10 @@ func (c *OutletConfiguration) Reset() {
 		Reporting:    reporter.DefaultConfiguration(),
 		Metadata:     metadata.DefaultConfiguration(),
 		Routing:      routing.DefaultConfiguration(),
-		Kafka:        kafka.DefaultConfiguration(),
+		Kafka:        kafkainput.DefaultConfiguration(),
 		ClickHouseDB: clickhousedb.DefaultConfiguration(),
 		ClickHouse:   clickhouse.DefaultConfiguration(),
-		KafkaOut:     kafkaout.DefaultConfiguration(),
+		KafkaOutput:  kafkaoutput.DefaultConfiguration(),
 		Flow:         flow.DefaultConfiguration(),
 		Core:         core.DefaultConfiguration(),
 		Schema:       schema.DefaultConfiguration(),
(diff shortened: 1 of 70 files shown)


common/schema/protobuf_test.go
```diff
@@ -0,0 +1,363 @@
...
+	if payload := bf.ProtobufMessage(); payload != nil {
+		t.Errorf("expected no Protobuf message when disabled, got %d bytes", len(payload))
+	}
+}
+
+// marshalFlowJSONReadback replicates solution 1 (the draft branch): it rebuilds
+// a flat JSON object by reading the last appended value of each ClickHouse
+// column. Kept here, test-only, purely to benchmark it head-to-head against the
+// append-time Protobuf encoder.
+func (bf *FlowMessage) marshalFlowJSONReadback() ([]byte, error) {
```
vincentbernat: This looks like a left over of another solution. This should be removed as it makes no sense to compare with a solution that were never there.

common/schema/protobuf_test.go
```diff
@@ -0,0 +1,363 @@
...
+	b.ReportAllocs()
+	b.ResetTimer()
+	for range b.N {
+		populateBenchFlow(bf)
+		bf.Finalize()
+		benchSink += len(bf.ProtobufMessage())
+	}
+}
+
+func BenchmarkEncodeJSONReadback(b *testing.B) {
```
vincentbernat: To be removed.

common/schema/protobuf_test.go
```diff
@@ -0,0 +1,363 @@
...
+		payload, err := bf.marshalFlowJSONReadback()
+		if err != nil {
+			b.Fatal(err)
+		}
+		benchSink += len(payload)
+		bf.Finalize()
+	}
+}
+
+func TestProtobufVsJSONSize(t *testing.T) {
```
vincentbernat: To be removed.

common/schema/message.go
vincentbernat: This needs to be renamed to `messageBatch` to match its new content.

common/schema/clickhouse.go
```diff
@@ -150,6 +150,9 @@ func (bf *FlowMessage) AppendDateTime(columnKey ColumnKey, value uint32) {
 	}
 	bf.batch.columnSet.Set(uint(columnKey))
 	col.(*proto.ColDateTime).AppendRaw(proto.DateTime(value))
+	if bf.batch.protobufEnabled {
```
vincentbernat: I feel like this check should be inside the function but we need to be sure the function is inlined. If it works, it is cleaner as the checks for `protobufEnabled` are all in `protobuf.go`.

cmd/akvorado/outlet.go
```diff
@@ -20,6 +20,7 @@ import (
 	"akvorado/outlet/core"
 	"akvorado/outlet/flow"
 	"akvorado/outlet/kafka"
+	"akvorado/outlet/kafkaout"
```
vincentbernat: Should be `kafkaoutput`.

cmd/akvorado/outlet.go
```diff
@@ -50,6 +52,7 @@ func (c *OutletConfiguration) Reset() {
 		Kafka:        kafka.DefaultConfiguration(),
 		ClickHouseDB: clickhousedb.DefaultConfiguration(),
 		ClickHouse:   clickhouse.DefaultConfiguration(),
+		KafkaOut:     kafkaout.DefaultConfiguration(),
```
vincentbernat: Should be `KafkaOutput` and `Kafka` should be renamed to `KafkaInput` since it's cleaner.

vincentbernat (changes requested):
I have already pushed some of the requested changes to your branch:

- moving checks if protobuf export is enabled into `protobuf.go` and ensuring the functions are inlined
- renaming kafka to kafkainput and kafkaout to kafkaoutput in both orchestrator and outlet (but I kept configuration name "kafka" in orchestrator)
- removal of tests/benchmarks for a previous "solution"

I did not update the changelog as the changes may need to be rebased, but we should let the user know about the metric name changes.

vincentbernat:
Wait for the tests to pass before trying anything. If some of them fail, I'll fix them and force push.

vincentbernat (approved):
I have pushed a last commit removing the extra tests. I am all good now. It's ready to merge if you are OK with it.

vincentbernat:
I have also pushed a few more changes:
- protobuf schema was not exposed like it was before Akvorado 2.0
- using TryProduce() for best-effort delivery to Kafka
- flushing Kafka records on shutdown
- make compression configurable (like for inlet)

neumayer:
Thanks, nice. Looks all cool at first glance, let me take a closer look and verify this all works with my current deployment (the protobuf format stuff will make using it much easier, and also the version bump story will be easier operationally). I'll report back.

neumayer:
Ok, updated plan:

* rebase my branch (did that and force pushed so I can do a clean rollout here from the fork)
* go through the full loop for my test setup just in case I missed some operational impact (can bump by new hash temporarily)
* analyse metrics and potential impact on parameters (I expect 24h to be enough now to feel confident)
* update pull request description based on new reality
* assuming all good I'll merge (what do you prefer here, looks like a mix of squash and rebase merges (asking because I'm not sure how easy the current history is to read, assuming you prefer rebase merge, that correct?)

TryProduce is a great improvement, what was I thinking not using that to start with :-)

neumayer:
My suggestion for the changelog would be:

+- 💥 *outlet*: rename the Kafka input metrics from `akvorado_outlet_kafka_*` to
+  `akvorado_outlet_kafkainput_*`, following the rename of the component; update
+  any dashboard or alert relying on them

(assuming you'll want to make the changelog changes yourself).

Mentioning because of course I ran into this myself when running the latest version :-)

vincentbernat:
You can add the changelog yourself. I didn't as it would trigger a conflict when rebasing. This is OK now.

As for the metrics, I was thinking about adding the possibility to alias them, but it means they will be stored twice. Not worth it I think.

As there are many commits, I'll do a merge commit. Some commits could be squashed together, but that's not worth it.

neumayer:
Pushed the changelog, and description is updated a bit.

I have preliminary good results from the real deployment (the rebased version without the changelog commit): looking good, basically no change from the version before your changes (so it's great with the simplifications).
Will verify again tomorrow just in case and post a short summary.

vincentbernat:
Let me know if you are OK to merge!

neumayer:
The reworked producer has now run ~48 h at production volume across two configurations, behaves just like before

We dropped the `queue-size` from 196608 to 65536 and removed our `load-balance: by-exporter` pin, so the second window runs on the upstream default partitioner:

| | `queue-size` / `load-balance` | duration | peak | drops | buffer peak |
|---|---|---|---|---|---|
| A | 196608, `by-exporter` | 27 h | 105k msg/s | **0** | 11,510 |
| B | 65536, `random` (default) | 21 h | **205k msg/s** | **0** | 16,491 (25% of cap) |

In window B we got a higher peak than in the last round of experiments — with zero drops and zero produce errors. This is consistent with the lower queue size we tried in the beginning, peak occupancy is slightly higher now. ClickHouse indistinguishable across windows.

On the `queue-size` default, since it is the only relevant setting: at ~100k msg/s steady state, buffer occupancy sat at ~2.7k records (p95), but the burst pushed it to 16.5k. So 4096 is comfortable at ordinary rates, and high-volume deployments will want more -- docs say this already.

This burst was a bit too much for our downstream consumer (confirmed it was a real input spike so no buffer related backfill behaviour), a Vector process on a 1 GiB limit that got OOM-killed. Confident this is our problem to work through, independent of the Akvorado changes (confirmed from our logs).

Nothing outstanding from our side — assuming you'll do the merge itself, thanks!

Looking forward to using this functionality from the next official release, and thanks for the help along the way, feeling good about this.

vincentbernat:
Great! Thanks for the PR and the tests!

barhom:
Is this meant to forward enriched packets to for example be used in ddos detectors?

neumayer:
@barhom this is to support kafka based detection on enriched flows, yes.

barhom:
> @barhom this is to support kafka based detection on enriched flows, yes.

I am very curious about this. Can you elaborate on what can detect ddos using Kafka encriched flows? 
I was looking into fastnetmon and it cannot as of yet use Kafka as input.

Would you say that fastnetmon would benefit from using Kafka encriched flows as input? Or do you suggest I look elsewhere for better detection?

neumayer:
I don't think I can comment on your use case.
The kafka outlet will put enriched flows into kafka, if you have any component of any kind that can use kafka as input then it might help you.

PR #2527, opened 2026-07-08
schema: add fixed-length /24 prefix dimensions (SrcNetPrefix24/DstNetPrefix24)

sebasmendez:
## What

Adds two computed dimensions, `SrcNetPrefix24` and `DstNetPrefix24`, that aggregate the source/destination **IPv4** address to its containing **/24**.

## Why

`SrcNetPrefix`/`DstNetPrefix` use a variable prefix length taken from the routing table, so the same traffic can spread across rows of different lengths (a `/22` aggregate, its more-specifics, etc.). For inter-domain traffic engineering the natural unit is the **/24 — the minimum BGP announcement/balancing unit**. Bucketing to a fixed /24 puts each /24 on a single row with its true total, so an operator can decide at a glance whether to deaggregate/rebalance a given /24.

This is particularly useful for a **transit AS whose destinations include downstream customers' own prefixes** (their ASNs/IPs), where a static `networks` list doesn't scale — the /24 bucket is agnostic to who owns the prefix.

## Implementation

- Computed **alias** column (no storage cost), mirroring the existing `NetPrefix` column but with the mask fixed to `24`.
- Only the `Src` column is defined; the `Dst` variant is generated automatically via the existing `Src -> Dst` expansion.
- IPv4 only (`EType = 0x800`); IPv6 yields an empty string.

Happy to generalize this into a **configurable prefix length** (and/or an IPv6 counterpart such as /48) if that direction is preferred — this is the minimal version to start the discussion.

### DIFF
--- a/common/schema/definition.go
+++ b/common/schema/definition.go
@@ -154,6 +154,8 @@ const (
 	ColumnDstNetMask
 	ColumnSrcNetPrefix
 	ColumnDstNetPrefix
+	ColumnSrcNetPrefix24
+	ColumnDstNetPrefix24
 	ColumnSrcAS
 	ColumnDstAS
 	ColumnSrcVlan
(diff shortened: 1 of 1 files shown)


vincentbernat:
Did you see that you can already do that in the interface? Check this [example](https://demo.akvorado.net/visualize/v1-N4IgRglgJhBOCmBjALhA9gOwIYBsQC4AzXAZ3gBoQYBbeDE9eggbRAGVZEBBKKWEALqU6UAiABMABnEA2ALSSA7AoAcAFQCMATnwBmFfgAsigHSHjALRCVCEHMnj98IAJIYXhAEJoArhihYsACeAAQAvCHwAB4OsNh4lADmsFgADgAWakGp8GIkyFiIANbwopTpPtRYGACi-mIYaADu1iAVVRhsBbDIYjIh6b6wJCFYiWitOBDUEL34GpKUUzPIWTliWABuia2pCJvoPiQACo7ookSkFCD5gXMS0vJKqpq6egbGZoYyVpTIsH5EFgHHJNoYCOJDH8ARggSDNjICIYVJQ-LMSGIcLowKkMQBfIA).

sebasmendez:
You're right, thanks! I missed the built-in IPv4/IPv6 truncation on the Src/Dst Addr dimensions — that does exactly what I needed (and handles IPv6 too), so this is redundant. Closing. Thanks for the pointer and for Akvorado!

PR #2548, opened 2026-07-30
outlet: move back GeoIP and networks into outlet

### LINKED ISSUE #2007: Move back GeoIP into outlet
### Disclaimer

- [x] I plan to implement the feature myself.

### Akvorado version

2.0.1

### Feature description

GeoIP should be moved back from ClickHouse to outlet.

vincentbernat:
We have more flexibility in the outlet, for example we can use networks for ASN before using routing component. This also do not require to build a giant CSV to be shipped to ClickHouse, which eats a lot of CPU and memory. On the other hand, with many GeoIP databases, we do more lookups and the behavior is slightly different. As documented in the changelog:

- 💥 *outlet*: the attributes from `outlet`→`networks` now win over GeoIP whatever their prefix length. Previously, both were merged into a single ClickHouse dictionary and the most specific prefix was winning: a country set on a subnet was overridden by the GeoIP country of a smaller one

cc @netixx (it's WIP, I need to work a bit more on it, wanted to let you know).

Fix #2007


vincentbernat:
I think it should be possible to get back the "unified" networks.csv that was previously there by merging all sources in a mmdb file that will be used for all lookups. One lookup would work for everything at once and, with the exception of a bit more code added to create that, the remaining code should be leaner.

vincentbernat:
I think I will in fact merge all sources into an in-memory table. So, it will use some memory (which can be swapped out) but it will keep the same semantics as previously.

vincentbernat:
So, like previously, a tree with all the network attributes is built in memory. On my workstation, with the demo, it takes less than 5 seconds to build and in the steady state, it uses about 200MB of memory. There are metrics for this:

```
# HELP akvorado_outlet_networks_memory_bytes Estimated memory taken by the prefixes and their attributes.
# TYPE akvorado_outlet_networks_memory_bytes gauge
akvorado_outlet_networks_memory_bytes 1.93659671e+08
# HELP akvorado_outlet_networks_prefixes Number of prefixes, including the ones from the GeoIP databases.
# TYPE akvorado_outlet_networks_prefixes gauge
akvorado_outlet_networks_prefixes 3.473796e+06
# HELP akvorado_outlet_networks_rebuild_last_time_seconds Time spent during the last rebuild of the networks.
# TYPE akvorado_outlet_networks_rebuild_last_time_seconds gauge
akvorado_outlet_networks_rebuild_last_time_seconds 4.479137755
```

It looks good enough for me and I'll merge it later.

PR #2565, opened 2026-08-05
outlet/metadata: cache negative results when all providers skip

safonov-nebius:
## Problem

  When an exporter is not known to any configured provider (no SNMP, no
  gNMI, no static entry), every Metadata.Lookup call for that exporter
  runs through the full provider loop, waits for the query deadline (up
  to QueryTimeout = 5 s, or InitialDelay = 1 min at startup), and returns
  empty — without caching the result.

  Because Lookup is called synchronously inside the Kafka consumer worker
  goroutine, and kgo.BlockRebalanceOnPoll() prevents AllowRebalance()
  until the entire batch is processed, a burst of flows from many unknown
  exporters can hold each worker blocked long enough that Kafka's heartbeat
  session expires. Once the heartbeat loop terminates (it stops after
  receiving REBALANCE_IN_PROGRESS), the broker removes the consumer from
  the group. The resulting rebalance is itself blocked by still-running
  workers, creating a cascade that stops all message consumption.

  Observed symptom: repeated REBALANCE_IN_PROGRESS / rejoin cycles every
  ~45 s (matching the default session.timeout.ms) with no partitions
  assigned, until the outlet stops consuming entirely.

  ## Fix

  When queryProviders exhausts all providers via ErrSkipProvider (all
  skipped, no transient error), store the negative result in a per-component
  sync.Map keyed by "ip-ifindex" with an expiry of NegativeCacheDuration
  (default 30 s). Lookup checks this map before entering the singleflight,
  returning immediately for known-unknown exporters without blocking the
  Kafka worker.

  - Expired entries are purged on the existing CacheCheckInterval ticker.
  - NegativeCacheDuration is configurable (1 s – 1 h, default 30 s).
  - Transient provider errors still go through the circuit breaker and are not cached.
  - A new TestNegativeCache test verifies the behaviour with fake time via synctest.

### DIFF
--- a/outlet/metadata/config.go
+++ b/outlet/metadata/config.go
@@ -32,6 +32,11 @@ type Configuration struct {
 	// InitialDelay defines how long to wait at start (when receiving the first
 	// packets) before applying the query timeout
 	InitialDelay time.Duration `validate:"min=1s,max=1h"`
+	// NegativeCacheDuration defines how long to cache a negative result when all
+	// providers skip a query (exporter unknown to any configured provider). This
+	// prevents repeated blocking lookups for exporters with no SNMP/gNMI source,
+	// which would otherwise hold up Kafka consumer workers.
+	NegativeCacheDuration time.Duration `validate:"min=1s,max=1h"`
 }
 
 // DefaultConfiguration represents the default configuration for the metadata provider.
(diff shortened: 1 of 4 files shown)


vincentbernat:
I am not 100% sure to understand your use case. There is already a negative cache (if an interface does not exist with the SNMP provider, it creates the entry as a zero entry). I would like to understand in which condition you do not run into this case.

vincentbernat:
Likely, you could keep the test and use:

```go
c.sc.Put(now, query, provider.Answer{})
```

At the end of the for loop. I don't quite see why `allSkipped` is needed. The only way to exit the loop is through the skip branch.

vincentbernat:
I have added a commit with my proposition.

vincentbernat:
This has been merged in e5ca8631f333.

PR #2576, opened 2026-08-11
Support for IPv6 literals on listen directives

lamaral:
While trying to force Akvorado to connect over IPv6 to ClickHouse or to listen only on IPv6 addresses, I encountered a problem with the configuration validation:

```
Error: invalid configuration:
Key: 'OrchestratorConfiguration.ClickHouseDB.Servers[0]' Error:Field validation for 'Servers[0]' failed on the 'listen' tag
Key: 'OrchestratorConfiguration.Outlet[0].ClickHouseDB.Servers[0]' Error:Field validation for 'Servers[0]' failed on the 'listen' tag
Key: 'OrchestratorConfiguration.Console[0].ClickHouse.Servers[0]' Error:Field validation for 'Servers[0]' failed on the 'listen' tag
```
when the clickhouse server list contained `"[::1]:9000"`

```
Error: invalid configuration:
Key: 'OrchestratorConfiguration.HTTP.Listen' Error:Field validation for 'Listen' failed on the 'listen' tag
```
when the listen directive was set to `"[::1]:8000"` and so on.

With this patch, I was able to get Akvorado to start when using IPv6 literals and everything seems to work fine.


vincentbernat:
Thanks!

PR #2582, opened 2026-08-12
orchestrator/clickhouse: add {database} to zookeeper path

wow0000:
hello!

This change add the macro {database} to the zookeeper path of replicated mergetrees. This allows multiple instances of Akvorado on a single ClickHouse cluster.

This pull request fix the following error when starting akvorado on a cluster with an already existing clickhouse schema:
```
5:01PM INF akvorado/orchestrator/clickhouse/migrations_helpers.go:221 > create dictionary protocols module=akvorado/orchestrator/clickhouse
5:01PM INF akvorado/orchestrator/clickhouse/migrations_helpers.go:221 > create dictionary icmp module=akvorado/orchestrator/clickhouse
5:01PM INF akvorado/orchestrator/clickhouse/migrations_helpers.go:221 > create dictionary tcp module=akvorado/orchestrator/clickhouse
5:01PM INF akvorado/orchestrator/clickhouse/migrations_helpers.go:221 > create dictionary udp module=akvorado/orchestrator/clickhouse
5:01PM ERR akvorado/orchestrator/clickhouse/root.go:93 > database migration error error="cannot create flows_local: code: 342, message: There was an error on [clickhouse-1:9000]: Code: 342. DB::Exception: Metadata of table toto2.flows_local (1674fa19-086f-4614-b639-60bc5fb8ae5e) in ZooKeeper differs in skip indexes. Stored in ZooKeeper: idx_exportername ExporterName TYPE minmax GRANULARITY 4, idx_srcaddr SrcAddr TYPE bloom_filter(0.001) GRANULARITY 4, idx_dstaddr DstAddr TYPE bloom_filter(0.001) GRANULARITY 4, idx_srcas SrcAS TYPE bloom_filter(0.001) GRANULARITY 4, idx_dstas DstAS TYPE bloom_filter(0.001) GRANULARITY 4, idx_srcport SrcPort TYPE bloom_filter(0.001) GRANULARITY 4, idx_dstport DstPort TYPE bloom_filter(0.001) GRANULARITY 4, idx_srccountry SrcCountry TYPE bloom_filter(0.001) GRANULARITY 4, idx_dstcountry DstCountry TYPE bloom_filter(0.001) GRANULARITY 4, idx_inifprovider InIfProvider TYPE set(0) GRANULARITY 4, idx_outifprovider OutIfProvider TYPE set(0) GRANULARITY 4, idx_inifconnectivity InIfConnectivity TYPE set(0) GRANULARITY 4, idx_outifconnectivity OutIfConnectivity TYPE set(0) GRANULARITY 4, idx_inifboundary InIfBoundary TYPE set(0) GRANULARITY 4, idx_outifboundary OutIfBoundary TYPE set(0) GRANULARITY 4, parsed from ZooKeeper: idx_exportername ExporterName TYPE minmax GRANULARITY 4, idx_srcaddr SrcAddr TYPE bloom_filter(0.001) GRANULARITY 4, idx_dstaddr DstAddr TYPE bloom_filter(0.001) GRANULARITY 4, idx_srcas SrcAS TYPE bloom_filter(0.001) GRANULARITY 4, idx_dstas DstAS TYPE bloom_filter(0.001) GRANULARITY 4, idx_srcport SrcPort TYPE bloom_filter(0.001) GRANULARITY 4, idx_dstport DstPort TYPE bloom_filter(0.001) GRANULARITY 4, idx_srccountry SrcCountry TYPE bloom_filter(0.001) GRANULARITY 4, idx_dstcountry DstCountry TYPE bloom_filter(0.001) GRANULARITY 4, idx_inifprovider InIfProvider TYPE set(0) GRANULARITY 4, idx_outifprovider OutIfProvider TYPE set(0) GRANULARITY 4, idx_inifconnectivity InIfConnectivity TYPE set(0) GRANULARITY 4, idx_outifconnectivity OutIfConnectivity TYPE set(0) GRANULARITY 4, idx_inifboundary InIfBoundary TYPE set(0) GRANULARITY 4, idx_outifboundary OutIfBoundary TYPE set(0) GRANULARITY 4, local: . (METADATA_MISMATCH) (version 26.3.17.110 (official build))" module=akvorado/orchestrator/clickhouse
```

Tests:
- Locally on the provided ClickHouse cluster OK
- make test OK
- Creating and filling the database before the patch then applying the patch OK


vincentbernat:
Unfortunately, it breaks the existing configurations.

wow0000:
Hello, 
> Unfortunately, it breaks the existing configurations.

My first testing was indeed wrong (probably due to a me error when rebuilding after switching from old to new)

I reverted the changes and added a configuration element that can allow the user to use a different zookeeper path when deploying a new instance of Akvorado.
As it is not automatic, I am not sure that this would be the best solution, however I did not find any "easy" changes to provide automatic support without adding a lot of code around it

Feel free to suggest changes or modification that would make this easier on the user and/or automatic for new database creations

The current commit allows configuration like this:
```yaml
clickhousedb:
  servers:
    - localhost:9001
  cluster: "akvorado"
  username: default
  password: ""
  database: akvorado2
  maxopenconns: 10
  dialtimeout: 5s
  zookeepersuffix: "/{database}"
  tls:
    enable: false
```

vincentbernat:
I think it's better to make a configuration setting to override the whole path (with a default to the current one).

wow0000:
Hello,

I made the requested changes, leaving the path up to the user with a identical default to the current one. I did use the `{table}` clickhouse macro instead of the table variable in the `mergeTreeEngine` function, tell me if you see any issue that could arise from this

orchestrator/clickhouse/migrations_helpers.go
```diff
@@ -167,10 +167,10 @@ func (c *Component) tableAlreadyExists(ctx context.Context, table, column string
 // Replicated if we are on a cluster.
 func (c *Component) mergeTreeEngine(table, variant string, args ...sb.Expr) sb.Engine {
 	if c.d.ClickHouse.ClusterName() != "" {
-		zkPath := fmt.Sprintf("/clickhouse/tables/shard-{shard}/%s%s", table, c.d.ClickHouse.ZooKeeperSuffix())
+		zkPath := c.d.ClickHouse.ZooKeeperPath()
 		if helpers.Testing() {
-			zkPath = fmt.Sprintf("/clickhouse/tables/shard-{shard}/%s/%s%s",
-				c.d.ClickHouse.DatabaseName(), table, c.d.ClickHouse.ZooKeeperSuffix())
+			zkPath = fmt.Sprintf("/clickhouse/tables/shard-{shard}/%s/%s",
+				c.d.ClickHouse.DatabaseName(), table)
```
vincentbernat: Move this condition in `DefaultConfiguration()`. This may require updating some of the tests.

config/akvorado.yaml
```diff
@@ -39,4 +39,4 @@ clickhouse:
 
 inlet: !include "inlet.yaml"
 outlet: !include "outlet.yaml"
-console: !include "console.yaml"
+console: !include "console.yaml"
```
vincentbernat: Missing newline.

vincentbernat (changes requested):

vincentbernat:
> I did use the `{table}` clickhouse macro instead of the table variable in the `mergeTreeEngine` function, tell me if you see any issue that could arise from this

I wonder what happens for existing installations because of that...

orchestrator/clickhouse/migrations_helpers.go
```diff
@@ -168,10 +167,9 @@ func (c *Component) tableAlreadyExists(ctx context.Context, table, column string
 func (c *Component) mergeTreeEngine(table, variant string, args ...sb.Expr) sb.Engine {
 	if c.d.ClickHouse.ClusterName() != "" {
 		zkPath := c.d.ClickHouse.ZooKeeperPath()
-		if helpers.Testing() {
-			zkPath = fmt.Sprintf("/clickhouse/tables/shard-{shard}/%s/%s",
-				c.d.ClickHouse.DatabaseName(), table)
-		}
+		zkPath = strings.ReplaceAll(zkPath, "{database}", c.d.ClickHouse.DatabaseName())
+		zkPath = strings.ReplaceAll(zkPath, "{table}", table)
```
vincentbernat: I think we need a comment explaining why we do that, otherwise people may start removing it. I think you only need to do that for `{table}`. Just say this is for backward compatibility.

vincentbernat (changes requested):
And ping me at the office to iterate faster.

orchestrator/clickhouse/migrations_helpers.go
vincentbernat: OK, I understand it breaks the tests if we keep `{database}`. So there are two reasons to expand both macros instead of letting ClickHouse do that:
- for `{table}`, it is to not trigger a migration for existing setups
- for `{database}`, it is to not trigger a migration for tests

I have put a proposition of comment.

orchestrator/clickhouse/migrations_helpers.go
```diff
@@ -167,11 +166,10 @@ func (c *Component) tableAlreadyExists(ctx context.Context, table, column string
 // Replicated if we are on a cluster.
 func (c *Component) mergeTreeEngine(table, variant string, args ...sb.Expr) sb.Engine {
 	if c.d.ClickHouse.ClusterName() != "" {
-		zkPath := fmt.Sprintf("/clickhouse/tables/shard-{shard}/%s", table)
-		if helpers.Testing() {
-			zkPath = fmt.Sprintf("/clickhouse/tables/shard-{shard}/%s/%s",
-				c.d.ClickHouse.DatabaseName(), table)
-		}
+		zkPath := c.d.ClickHouse.ZooKeeperPath()
```
vincentbernat:
```suggestion
		zkPath := c.d.ClickHouse.ZooKeeperPath()
		// Expand some of the built-in substitutions for backward compatibility
```

vincentbernat:
Another idea. Instead of hardcoding a default value, we could keep the empty value as a way to autodetect.

- `zookeeper-path` has a value, use it as is (like currently)
- `zookeeper-path` has no value and the table exists, extract the value in the table
- `zookeeper-path` has no value and the table does not exist, use the value with `{database}`
- `zookeeper-path` defaults to the empty value

Or as it does not seem possible to change this value once running:

- the table exists, use the current value instead of `zookeeper-path`
- the table does not exist, use the value of `zookeeper-path`
- `zookeeper-path` defaults to the value with `{database}` in it

This way, we don't break existing setups (prod or tests) and for new setups, we are using the new safer value. I say that because other people got bit by this (see #2597).

wow0000:
Hello Vincent,

I implemented the second possibility that you offered and removed ZookeeperPath so that if we ever need to similar modifications, upgrade/changes would be possible and easier without any user intervention (matching on either of the two version of possible zkpath).

It seems to me that, by including `{database}`, `{table}` and `{shard}` in the path, we cover all cases when having different Akvorado instances on the same (Zoo)Keeper leaving not many reasons to make it configurable, tell me what you think about this

This change makes deployment of new instances possible and should not break previous installations

vincentbernat (changes requested):
The idea was to keep the exact path when the table exists and only use the new path when it does not.

orchestrator/clickhouse/migrations_helpers.go
```diff
@@ -163,15 +165,25 @@ func (c *Component) tableAlreadyExists(ctx context.Context, table, column string
 	return false, nil
 }
 
+func (c *Component) queryTableEngine(ctx context.Context, table string) string {
```
vincentbernat:
```suggestion
func (c *Component) queryExistingZkPath(ctx context.Context, table string) string {
```

orchestrator/clickhouse/migrations_helpers.go
```diff
@@ -163,15 +165,25 @@ func (c *Component) tableAlreadyExists(ctx context.Context, table, column string
...
+		subMatches := replicatedMergeTreeRegexp.FindStringSubmatch(createQuery)
+		if len(subMatches) == 2 {
+			return subMatches[1]
+		}
+	}
+	return fmt.Sprintf("/clickhouse/tables/shard-{shard}/%s/%s", c.d.ClickHouse.DatabaseName(), table)
+}
+
 // mergeTreeEngine returns a MergeTree engine definition, either plain or using
 // Replicated if we are on a cluster.
```
vincentbernat:
```suggestion
// Replicated if we are on a cluster. Zookeeper path from an existing table is kept unchanged.
```

orchestrator/clickhouse/migrations_helpers.go
```diff
@@ -163,15 +165,25 @@ func (c *Component) tableAlreadyExists(ctx context.Context, table, column string
...
-func (c *Component) mergeTreeEngine(table, variant string, args ...sb.Expr) sb.Engine {
+func (c *Component) mergeTreeEngine(ctx context.Context, table, variant string, args ...sb.Expr) sb.Engine {
 	if c.d.ClickHouse.ClusterName() != "" {
-		zkPath := fmt.Sprintf("/clickhouse/tables/shard-{shard}/%s", table)
-		if helpers.Testing() {
-			zkPath = fmt.Sprintf("/clickhouse/tables/shard-{shard}/%s/%s",
-				c.d.ClickHouse.DatabaseName(), table)
-		}
+		zkPath := c.queryTableEngine(ctx, table)
```
vincentbernat:
```suggestion
		zkPath := c.queryExistingZkPath(ctx, table)
```

vincentbernat (changes requested):
Also, we could use the SQL parser to extract the zookeeper path without using a regex. But I don't want to annoy you too much and this can be done in a second step (and this could be applied for settings as well).

vincentbernat:
Thanks!

PR #2583, opened 2026-08-12
outlet/metadata: expose when no provider has metadata for an exporter

arunrajm91:
When every configured provider skips an exporter (no SNMP, no gNMI, no matching static entry), the flow is enriched with empty metadata and, since e5ca863, the negative result is also cached — with no trace anywhere: no log line and no metric.

Debugging this from the outside means staring at flows with empty interface names and guessing which layer dropped them. I recently lost a few hours to exactly this while pointing a device at the SNMP provider when the device could not answer ifXTable queries (SonicWall — it has no ifXTable at all): nothing anywhere said which exporter was failing to resolve, or why.

This PR adds:

- a `provider_skips_total` counter labelled with the exporter address — same pattern and cardinality bound as the existing `provider_breaker_opens_total`;
- a rate-limited warning log naming the exporter.

With this, an unmatched exporter is visible at a glance in metrics and logs. Includes a test following the style of the recent negative-caching test.

### DIFF
--- a/outlet/metadata/root_test.go
+++ b/outlet/metadata/root_test.go
@@ -361,20 +361,25 @@ func TestNegativeCache(t *testing.T) {
 		configuration.Providers = []ProviderConfiguration{{Config: skipProviderConfiguration{}}}
 		c := NewMock(t, r, configuration, Dependencies{Daemon: daemon.NewMock(t)})
 
-		// First lookup: no cached result, provider is queried and skips.
+		// First lookup: no cached result, provider is queried and skips. The
+		// skip is visible in metrics, labelled with the exporter address.
 		expectMockLookup(t, c, "127.0.0.1", 765, provider.Answer{})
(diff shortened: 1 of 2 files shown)


outlet/metadata/root_test.go
```diff
@@ -389,3 +389,31 @@ func TestNegativeCache(t *testing.T) {
 		}
 	})
 }
+
+func TestProviderSkipsMetric(t *testing.T) {
```
vincentbernat: You can merge this test with the previous one, since this tests the same logic, just a different effect.

vincentbernat (changes requested):

outlet/metadata/root_test.go
arunrajm91: Merged it into `TestNegativeCache` — it now asserts `requests_total` and `skips_total` together at each stage, which also covers that a cached negative answer does not count as another skip.

vincentbernat:
Thanks!

PR #2580, opened 2026-08-12
outlet/bmp: flush peer from shards concurrently

bogi788:
I've been troubleshooting an issue with outlets falling behind with processing flows when peers are removed.
The effected outlets usually have 600-950 peers and 100M to 150M routes.

I've come up with this PR as a workaround:
It adds an alternative version of `FlushPeer(...)` that processes all shards concurrently using goroutines.
This new implementation is only called if at least 4 peers and 512 * <number of shards> prefixes are present in order to avoid the synchronisation overhead for small setups.

It further changes the prefix tree "cleanup" to delete removed prefixes instead of rebuilding it entirely if there were at least 4 peers before flushing.

This seems to reduce the time for flushing a peer considerably with many peers and routes.
First the benchmark results for the current main branch (up to 1M routes and 50 peers):

```
BenchmarkRIBFlush/1000_routes,_1_peers-12         	    4776	         0.2364 ms/op
BenchmarkRIBFlush/1000_routes,_2_peers-12         	    1758	         0.6391 ms/op
BenchmarkRIBFlush/1000_routes,_5_peers-12         	     854	         1.317 ms/op
BenchmarkRIBFlush/1000_routes,_10_peers-12        	     348	         3.324 ms/op
BenchmarkRIBFlush/1000_routes,_20_peers-12        	     200	         5.802 ms/op
BenchmarkRIBFlush/1000_routes,_50_peers-12        	      70	        17.59 ms/op
BenchmarkRIBFlush/10000_routes,_1_peers-12        	     428	         2.630 ms/op
BenchmarkRIBFlush/10000_routes,_2_peers-12        	     188	         6.326 ms/op
BenchmarkRIBFlush/10000_routes,_5_peers-12        	      73	        17.55 ms/op
BenchmarkRIBFlush/10000_routes,_10_peers-12       	      26	        44.48 ms/op
BenchmarkRIBFlush/10000_routes,_20_peers-12       	      10	       100.4 ms/op
BenchmarkRIBFlush/10000_routes,_50_peers-12       	       5	       245.2 ms/op
BenchmarkRIBFlush/100000_routes,_1_peers-12       	      24	        47.43 ms/op
BenchmarkRIBFlush/100000_routes,_2_peers-12       	       9	       113.6 ms/op
BenchmarkRIBFlush/100000_routes,_5_peers-12       	       4	       272.7 ms/op
BenchmarkRIBFlush/100000_routes,_10_peers-12      	       3	       476.4 ms/op
[...]

PR #2594, opened 2026-08-17
[draft] outlet/bmp: flush removed peers in batches

bogi788:
This version of the PR depends on https://github.com/akvorado/akvorado/pull/2580. https://github.com/bogi788/akvorado/pull/1/changes shows only the changes on top.

As described in https://github.com/akvorado/akvorado/discussions/2587 BMP connections dropping and reconnecting leads to multiple peers becoming stale and subsequently being removed concurrently, which can cause a delay with inserting flows into clickhouse.

To work around this issue we've implemented batching of peer removals - both from BMP messages as well as from becoming stale - by accumulating peers over a ten second window before flushing them.

### DIFF
--- a/outlet/routing/provider/bmp/events.go
+++ b/outlet/routing/provider/bmp/events.go
@@ -108,7 +108,7 @@ func (p *Provider) removePeer(pkey peerKey, reason string) {
 	if !ok {
 		return
 	}
-	routesRemoved, prefixesRemoved := p.rib.FlushPeer(pinfo.reference)
+	routesRemoved, prefixesRemoved := p.rib.FlushPeer(pinfo.reference, len(p.peers))
 	delete(p.peers, pkey)
 	p.metrics.routes.WithLabelValues(exporterStr).Sub(float64(routesRemoved))
 	p.metrics.prefixesRemoved.WithLabelValues(exporterStr).Add(float64(prefixesRemoved))
--- a/outlet/routing/provider/bmp/events.go
+++ b/outlet/routing/provider/bmp/events.go
@@ -249,6 +249,8 @@ func (p *Provider) handlePeerUpNotification(pkey peerKey, body *bmp.BMPPeerUpNot
 		if !p.cancelRemovePeer(pkey) {
 			p.r.Info().Msgf("received extra peer up from exporter %s for peer %s",
 				exporterStr, peerStr)
+		} else {
+			pinfo.staleUntil = time.Time{}
 		}
 	} else {
 		// Peer does not exist at all
@@ -380,7 +382,9 @@ func (p *Provider) handleRouteMonitoring(pkey peerKey, body *bmp.BMPRouteMonitor
 
 	p.mu.Lock()
 	pinfo, ok := p.peers[pkey]
-	if !ok {
+	if ok {
+		p.cancelRemovePeer(pkey)
+	} else {
 		// We may have missed the peer down notification?
 		p.r.Info().Msgf("received route monitoring from exporter %s for peer %s, but no peer up",
 			exporterStr, peerStr)
--- a/outlet/routing/provider/bmp/events.go
+++ b/outlet/routing/provider/bmp/events.go
@@ -32,6 +32,12 @@ type peerInfo struct {
 	marshallingOptions []*bgp.MarshallingOption // decoding option (add-path mostly)
 }
 
+// peerToRemove describes a peer queued for batched removal.
+type peerToRemove struct {
+	reference   uint32 // reference of the peer in the RIB when it was queued
+	exporterStr string // exporter the peer belongs to, for metrics
+}
+
 // peerKeyFromBMPPeerHeader computes the peer key from the BMP peer header.
 func peerKeyFromBMPPeerHeader(exporter netip.AddrPort, header *bmp.BMPPeerHeader) peerKey {
 	return peerKey{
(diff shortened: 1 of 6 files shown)


bogi788:
Adapting `FlushPeer(...)` to accept a "set" of peer keys seems to have a tiny effect on performance, but overall it still seems to be faster than FlushPeer without https://github.com/akvorado/akvorado/pull/2580:

```
BenchmarkRIBFlush/1000_routes,_1_peers-12         	    5708	         0.2094 ms/op
BenchmarkRIBFlush/1000_routes,_2_peers-12         	    1950	         0.6324 ms/op
BenchmarkRIBFlush/1000_routes,_5_peers-12         	    1348	         0.9193 ms/op
BenchmarkRIBFlush/1000_routes,_10_peers-12        	     490	         2.353 ms/op
BenchmarkRIBFlush/10000_routes,_1_peers-12        	     763	         1.487 ms/op
BenchmarkRIBFlush/10000_routes,_2_peers-12        	     231	         5.286 ms/op
BenchmarkRIBFlush/10000_routes,_5_peers-12        	      90	        13.15 ms/op
BenchmarkRIBFlush/10000_routes,_10_peers-12       	      49	        25.28 ms/op
BenchmarkRIBFlush/100000_routes,_1_peers-12       	      86	        15.06 ms/op
BenchmarkRIBFlush/100000_routes,_2_peers-12       	      18	        60.41 ms/op
BenchmarkRIBFlush/100000_routes,_5_peers-12       	      10	       113.1 ms/op
BenchmarkRIBFlush/100000_routes,_10_peers-12      	       7	       165.4 ms/op
```

After running for a couple of hours this seems to result in roughly 3 peers removed per call to `FlushPeer`:

image

We've also seen a case where 96 peers were removed at once:

image

(This would likely have led to the automation restarting the outlet to prevent the delay in flows being processed to become too high without this PR.)

bogi788:
I believe this visualisation is well suited to showing the effects of this PR:

image

Its possible to tell when I rolled out the patch on our outlets by looking at it.

bogi788:
We rebased this PR and implemented two versions of the `shouldRemove` function (passed to `removeRoutes`) in order to reduce the overhead when removing few peers:

```
BenchmarkRIBFlush/1000_routes,_1_peers-12         	    5960	         0.2013 ms/op
BenchmarkRIBFlush/1000_routes,_2_peers-12         	    2222	         0.5252 ms/op
BenchmarkRIBFlush/1000_routes,_5_peers-12         	    1386	         0.8868 ms/op
BenchmarkRIBFlush/1000_routes,_10_peers-12        	     519	         2.317 ms/op
BenchmarkRIBFlush/1000_routes,_20_peers-12        	     379	         3.243 ms/op
BenchmarkRIBFlush/10000_routes,_1_peers-12        	     958	         1.249 ms/op
BenchmarkRIBFlush/10000_routes,_2_peers-12        	     284	         4.257 ms/op
BenchmarkRIBFlush/10000_routes,_5_peers-12        	      99	        12.85 ms/op
BenchmarkRIBFlush/10000_routes,_10_peers-12       	      44	        24.65 ms/op
BenchmarkRIBFlush/10000_routes,_20_peers-12       	      22	        48.19 ms/op
BenchmarkRIBFlush/100000_routes,_1_peers-12       	      86	        13.81 ms/op
BenchmarkRIBFlush/100000_routes,_2_peers-12       	      27	        48.28 ms/op
BenchmarkRIBFlush/100000_routes,_5_peers-12       	       9	       112.3 ms/op
BenchmarkRIBFlush/100000_routes,_10_peers-12      	       7	       159.6 ms/op
BenchmarkRIBFlush/100000_routes,_20_peers-12      	       5	       255.8 ms/op
BenchmarkRIBFlush/1000000_routes,_1_peers-12      	       7	       147.5 ms/op
BenchmarkRIBFlush/1000000_routes,_2_peers-12      	       3	       422.6 ms/op
BenchmarkRIBFlush/1000000_routes,_5_peers-12      	       2	       621.0 ms/op
BenchmarkRIBFlush/1000000_routes,_10_peers-12     	       2	       972.7 ms/op
BenchmarkRIBFlush/1000000_routes,_20_peers-12     	       1	      1682 ms/op
```

vincentbernat:
I didn't have a look yet, but I was thinking about the global problem. If you have 500 peers, it means there is one tree for 500 peers and I suppose there are many prefixes common to all these peers. For these prefixes, there is a slice with the 500 paths, one for each peer. The outlet will walk them one by one to match the next hop, this looks quite suboptimal.

I want to support the use case where you don't need to configure all the exporters to send their routes. But if it was not the case, we could have one tree per exporter, and a lot of problems would go away, but there is no easy fallback if an exporter is missing in this case. And it increases memory usage, depending on how much we share and don't share. For performance, we could share nothing, so each exporter has its tree and its interned attributes. This would solve a lot of problems for you. First, would the memory trade-off be OK? It would also be more exact, because currently, if one peer has a more specific, it will be used.

The downside of this is that we have to maintain two code paths, but maybe this could be a reasonable trade-off. We could also take a middle road: N exporters per tree. In fact, now I said it, I quite like the idea. You could use N=1, N=4 or the special value N=0 for the current implementation.

bogi788:
In our current setup requiring a tree for each exporter would be OK - we have BMP configured on all routers. The exporter addresses for flows and routing information (as seen from akvorado) doesn't match however because all BMP traffic is sent over encrypted tunnels from the inlets' VMs to the outlets.

Getting additional RAM assigned to the Outlets' VM could be tricky at the moment, depending on how much extra would be needed.

To give an update on the branch: Since the last update to the branch 4 days ago 10 peers were removed per `FlushPeer(...)` call on average:

image

Number of peers and routes for scale:

image

image

PR #2605, opened 2026-08-25
feat/outlet: Configurable alternative for workers to write in Clickhouse nodes

dimbdr:
Following https://github.com/akvorado/akvorado/discussions/2595

Use case: with the current way workers are writing to Clickhouse, I have often imbalance in writings with a node receiving significantly more than other nodes.

The goal of this PR is to propose another way to configure the connection between outlet's workers and Clickhouse nodes.
Instead of one worker sticking to one server and send all its batches to this same server, the alternate configuration allows a worker to send batches to all available Clickhouse servers in a round-robin manner.

Following a capture of a Clickhouse metric demonstrating the inserted rows done by the outlet. Each arrow indicates a restart of the outlet.
image
After each restart, writings remain balanced across the nodes.

By default when outlet is deployed, the workers are configured as current: "sticky" to a single Clickhouse server.
The "round-robin" way is configurable in outlet configuration.


outlet/clickhouse/config.go
```diff
@@ -7,6 +7,18 @@ import (
...
+const (
+	// ServerSelectionStickyRandom pins each worker to a single, randomly-chosen
+	// ClickHouse server for the lifetime of its connection. A new server is
+	// picked (again at random) only when the connection breaks. This is the
+	// legacy behavior and the default.
+	ServerSelectionStickyRandom = "sticky-random"
+	// ServerSelectionRoundRobin spreads a worker's batches across all configured
+	// ClickHouse servers in round-robin order so insert load is balanced across
+	// nodes.
+	ServerSelectionRoundRobin = "round-robin"
```
vincentbernat: It should be its own type: `ServerSelectionAlgorithm`. Then, you can use enumer to get automatic translation from and to string. There are many examples like this in the code. Look at the `Makefile` for integration.

console/data/docs/50-configuration.md
```diff
@@ -859,19 +859,30 @@ GeoIP database, like a city-level one, this uses a significant amount of memory.
...
 The default value is 100 000 and allows ClickHouse to handle incoming flows
 efficiently.
 
+When several ClickHouse servers are configured, `server-selection` controls how
+each worker spreads its batches across them:
+
+- `sticky-random` (the default) pins each worker to a single, randomly-chosen
+  server and keeps using it as long as the connection stays healthy. A new
+  server is picked (again at random) only when the connection breaks. This is
+  the historical behavior.
```
vincentbernat: The last sentence does not bring anything (all the more this is the default).

outlet/clickhouse/config.go
```diff
@@ -15,6 +27,9 @@ type Configuration struct {
 	MaximumBatchSize uint `validate:"min=1"`
 	// MaximumWaitTime is the maximum number of seconds to wait before sending the current batch.
 	MaximumWaitTime time.Duration `validate:"min=100ms"`
+	// ServerSelection controls how each worker chooses a ClickHouse server for a
+	// batch: "sticky-random" (default, legacy) or "round-robin".
+	ServerSelection string `validate:"oneof=sticky-random round-robin"`
```
vincentbernat: And use the new type here.

outlet/clickhouse/functional_test.go
```diff
@@ -21,6 +21,20 @@ import (
 )
 
 func TestInsert(t *testing.T) {
+	// Run the same scenario under both server-selection strategies. With a single
+	// server they behave identically (metrics exclude the per-server batch count),
+	// so the assertions hold for both while covering each selection branch.
```
vincentbernat: Better than nothing, but this does not help a lot.

outlet/clickhouse/worker.go
```diff
@@ -38,29 +38,59 @@ const (
...
+	// connectFn establishes/validates the connection for a server. It defaults to
+	// ensureConnected and is a field so tests can inject a fake dialer.
+	connectFn func(context.Context, *serverConn) error
+	// shuffleFn returns a random permutation of [0,n) and is used by the
+	// sticky-random strategy to pick a server. It defaults to rand.Perm and is a
+	// field so tests can make the choice deterministic.
+	shuffleFn     func(int) []int
 	options       ch.Options
 	asyncSettings []ch.Setting
 }
```
vincentbernat, line 72: Instead of having fields that work for some strategy or another, I would suggest to have each strategy its own worker (`stickyRandomWorker` or `roundRobinWorker`). Likely, a lot of code can be reused between them but I think the separation will be cleaner. You can also embed a `commonWorker` to store anything not specific to a strategy (attributes and methods).

outlet/clickhouse/worker_internal_test.go
```diff
@@ -0,0 +1,231 @@
...
+func TestPickServerStaggeredStart(t *testing.T) {
+	ok := func(context.Context, *serverConn) error { return nil }
+	// A worker whose cursor starts at index 1 (staggered by worker id) should
+	// begin on s1, not s0.
+	w := newTestWorker([]string{"s0", "s1", "s2"}, ok)
+	w.next = 1
+
+	got := pickN(t, w, 3)
+	want := []string{"s1", "s2", "s0"}
+	if !equalStrings(got, want) {
```
vincentbernat: Use `helpers.Diff`.

outlet/clickhouse/worker_internal_test.go
```diff
@@ -0,0 +1,231 @@
...
+	return got
+}
+
+func TestPickServerRoundRobin(t *testing.T) {
+	ok := func(context.Context, *serverConn) error { return nil }
+	w := newTestWorker([]string{"s0", "s1", "s2"}, ok)
+
+	got := pickN(t, w, 7)
+	want := []string{"s0", "s1", "s2", "s0", "s1", "s2", "s0"}
+	if !equalStrings(got, want) {
```
vincentbernat: Use `helpers.Diff`.

outlet/clickhouse/worker_internal_test.go
```diff
@@ -0,0 +1,231 @@
...
+		return nil
+	}
+	w := newTestWorker([]string{"s0", "s1", "s2"}, connectFn)
+
+	sc, err := w.pickServer(context.Background())
+	if err != nil {
+		t.Fatalf("pickServer() error: %v", err)
+	}
+	if sc.address != "s1" {
+		t.Errorf("expected failover to s1, got %s", sc.address)
```
vincentbernat:
```go
if diff := helpers.Diff(sc.address,` "s1"); diff != "" {
  t.Errorf("pickServer() (-got, +want):\n%s", diff)
}
```

outlet/clickhouse/worker_internal_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+func TestPickServerStickyStaysPinned(t *testing.T) {
+	ok := func(context.Context, *serverConn) error { return nil }
+	w := newTestWorker([]string{"s0", "s1", "s2"}, ok)
+	// The random pick lands on s2 first; the worker must then stay on it.
+	w.shuffleFn = fixedShuffle(2, 0, 1)
+
+	got := stickyPickN(t, w, 5)
+	want := []string{"s2", "s2", "s2", "s2", "s2"}
+	if !equalStrings(got, want) {
```
vincentbernat: You get the gist: `helpers.Diff`.

outlet/clickhouse/worker_internal_test.go
```diff
@@ -0,0 +1,231 @@
...
+		bf:     bf,
+		logger: r.With().Logger(),
+		selectServer: func(context.Context) (*serverConn, error) {
+			return nil, errors.New("no server available")
+		},
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+	defer cancel()
+	w.Flush(ctx)
```
vincentbernat, line 188: The test is missing. Just coverage is not enough. Maybe you could test the metrics?

outlet/clickhouse/worker.go
```diff
@@ -38,29 +38,59 @@ const (
...
 type realWorker struct {
 	c      *realComponent
 	bf     *schema.FlowMessage
 	last   time.Time
 	logger reporter.Logger
 
-	conn          *ch.Client
-	servers       []string
+	// servers holds one entry per configured ClickHouse server.
+	servers []*serverConn
```
vincentbernat, line 63: It was already the case before, but it is now worse. When a worker is shutdown (during a scale down for example), the connection was not closed. Here, the situation is now worse.

outlet/clickhouse/functional_test.go
```diff
@@ -111,7 +126,7 @@ func TestInsert(t *testing.T) {
 		}
 
 		// Check metrics
-		gotMetrics := r.GetMetrics("akvorado_outlet_clickhouse_", "-insert_time", "-wait_time")
+		gotMetrics := r.GetMetrics("akvorado_outlet_clickhouse_", "-insert_time", "-wait_time", "-batches_sent")
```
vincentbernat: You could keep `batches_sent`, couldn't you?

vincentbernat (changes requested):

outlet/clickhouse/config.go
dimbdr: Done, changed to enum in 2920a2aadc90837ef79fa9a198233d219ae12146

console/data/docs/50-configuration.md
dimbdr: Removed

outlet/clickhouse/worker.go
dimbdr: Donein d5cc9d57a27065c50b79352afaf4b4929fabb952 , threre's a `commonWorker` that can be specialized in 2 kind of workers now

outlet/clickhouse/worker_internal_test.go
dimbdr: Done in 048c981af6e4dcc198a4f39d6ac94ee9f6a713db

outlet/clickhouse/worker_internal_test.go
dimbdr: Done in 048c981af6e4dcc198a4f39d6ac94ee9f6a713db

outlet/clickhouse/worker_internal_test.go
dimbdr: Done

outlet/clickhouse/worker_internal_test.go
dimbdr: Done

outlet/clickhouse/worker.go
dimbdr: Indeed, connections were leaking when worker is shutdown.
This is managed in d82187b193eaf30141173e27f6605f0fb07fa33e

outlet/clickhouse/functional_test.go
dimbdr: Kept it, and tested the metric

dimbdr:
Ups sorry btw, I pushed the branch with a rebase to main, ungracefully

outlet/clickhouse/worker.go
```diff
@@ -22,6 +22,8 @@ import (
 type Worker interface {
 	FinalizeAndSend(context.Context) WorkerStatus
 	Flush(context.Context)
+	// Close releases the connections held by the worker.
```
vincentbernat: Useless comment, the method name is self-descriptive (and it's odd to add a comment for this one and not for the others).

outlet/clickhouse/worker.go
```diff
@@ -168,64 +202,147 @@ func (w *realWorker) Flush(ctx context.Context) {
...
-			w.c.metrics.errors.WithLabelValues("ping").Inc()
-			conn.Close()
-			conn = nil
+// FinalizeAndSend implements Worker.
+func (w *stickyRandomWorker) FinalizeAndSend(ctx context.Context) WorkerStatus {
+	return w.finalizeAndSend(ctx, w.pickServer)
+}
+
+// Flush implements Worker.
+func (w *stickyRandomWorker) Flush(ctx context.Context) { w.flush(ctx, w.pickServer) }
```
vincentbernat: Both implementations have the same `FinalizeAndSend()` and `Flush()` methods. Likely, we could put `pickServer()` method into `commonWorker` structure like `connectFn`.

outlet/clickhouse/worker_internal_test.go
```diff
@@ -0,0 +1,240 @@
...
+	if got := pickN(t, w.pickServer, 1); got[0] != "s0" {
+		t.Fatalf("expected initial pin to s0, got %s", got[0])
+	}
+	down["s0"] = true
+
+	// The reuse attempt fails, so the worker drops the pin and re-picks the next
+	// healthy server.
+	sc, err := w.pickServer(context.Background())
+	if err != nil {
+		t.Fatalf("pickServer() error: %v", err)
```
vincentbernat:
```suggestion
		t.Fatalf("pickServer() error:\n%v", err)
```

outlet/clickhouse/worker_internal_test.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// pickN calls pick n times and returns the sequence of selected addresses.
+func pickN(t *testing.T, pick func(context.Context) (*serverConn, error), n int) []string {
+	t.Helper()
+	got := make([]string, 0, n)
+	for range n {
+		sc, err := pick(context.Background())
+		if err != nil {
+			t.Fatalf("pickServer() error: %v", err)
```
vincentbernat:
```suggestion
			t.Fatalf("pickServer() error:\n%v", err)
```

outlet/clickhouse/worker_internal_test.go
```diff
@@ -0,0 +1,240 @@
...
+		if sc.address == "s0" {
+			return errBoom
+		}
+		return nil
+	}
+	w := newRoundRobin([]string{"s0", "s1", "s2"}, connectFn)
+
+	sc, err := w.pickServer(context.Background())
+	if err != nil {
+		t.Fatalf("pickServer() error: %v", err)
```
vincentbernat:
```suggestion
		t.Fatalf("pickServer() error:\n%v", err)
```

outlet/clickhouse/functional_test.go
vincentbernat, line 248: This seems a good place to add a test for `Close()`.

vincentbernat (changes requested):
Almost there!

outlet/clickhouse/worker.go
dimbdr: Removed

outlet/clickhouse/worker.go
dimbdr: Indeed, this is done in 9019948ef07e2eb35b1199bbeb7df268e7044c98

outlet/clickhouse/worker_internal_test.go
dimbdr: Done

outlet/clickhouse/worker_internal_test.go
dimbdr: Done

outlet/clickhouse/worker_internal_test.go
dimbdr: Done

outlet/clickhouse/functional_test.go
dimbdr: Added in d9fcd7b1acbbdca2c1d86284dd24f8683a9e0c4e

vincentbernat:
I have pushed two other commits for the remaining issues:
- `TestClose()` was testing the wrong thing and relying on ClickHouse server, which could run several tests in parallel
- `setup()` function was too tiny and generic to warrant its existence.

I'll merge once the tests are OK.

PR #2607, opened 2026-08-26
outlet/geoip: read db-ip database

dimbdr:
Hello :wave: 

Goal of this PR is to be able to read [DB-IP](https://db-ip.com/) databases, to enrich our data with their geoip info.
The small difference is that DB-IP provides `subdivisions[0].names.en ("California", "Saxony")` instead of iso_code like in Maxmind, so we would get only city and country but no state. With this change we can correctly read DB-IP database.

As for the use case, Maxmind Lite version lacks some info in city and country fields, whereas in DB-IP they are present. Following an example:

Maxmind
image

DB-IP
image


console/data/docs/50-configuration.md
```diff
@@ -806,6 +806,14 @@ following keys:
 
 [MaxMind DB file format]: https://maxmind.github.io/MaxMind-DB/
 
+The database provider is detected from the file metadata. MaxMind (GeoLite2 /
+GeoIP2), [IPinfo][] and [DB-IP][] databases are supported. DB-IP databases do
+not carry subdivision ISO codes, so the subdivision (state) is taken from its
+English name (e.g. `California`, `Saxony`) instead.
```
vincentbernat: The second sentence is useless for a user. Remove it.

outlet/geoip/iter_dbip.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return mmdbdata.Cursor{}, err
+	}
+	next := entries.First()
+	for range entries.Len() {
+		key, value, keyErr := next.ReadMapKey()
+		if keyErr != nil {
+			return mmdbdata.Cursor{}, keyErr
+		}
+		if string(key) == "names" {
```
vincentbernat: Since it is the first difference, I would suggest to declare that DBIP databases are maxmind ones, use `iso_code` if available and `names.en` as a fallback. This would avoid a lot of code duplication.

vincentbernat (changes requested):

console/data/docs/50-configuration.md
dimbdr: Removed

outlet/geoip/iter_dbip.go
dimbdr: Done

vincentbernat (approved):

vincentbernat:
Thanks!

PR #2631, opened 2026-09-13
Add option to select the timezone in the UI

nitzan-tz:
Curently the timezone is the browser timezone but in some cases when you work with a team in multiple timezones it is required to have a common timezone.
This option mimics similar option in Grafana


vincentbernat:
Currently, this is far too much code for such a feature. It spreads too much.

console/frontend/src/composables/useTimezone.ts
```diff
@@ -0,0 +1,92 @@
...
+  "akvorado-timezone",
+  "browser",
+);
+
+export const isBrowser = computed(() => selectedTimezone.value === "browser");
+export const isUTC = computed(() => selectedTimezone.value === "UTC");
+
+// Pre-computed list of timezone options
+export const timezones: TimezoneItem[] = [
+  { id: 1, code: "browser", name: `Browser Time (${browserTimezone})` },
```
vincentbernat: This makes the field big for most users. Just `Browser time`.

vincentbernat:
The change size is now OK. Thanks!

- The dates displayed in the start/end fields are still local to the browser time zone. It's confusing. I don't know if there is an easy way around that.
- The X-axis tooltip stays in local browser time.
- The tick for the day disappear. Dunno exactly why either, but previously, midnight was converted to the day. If we zoom out for several days, the X ticks become a bit odd.
- Just "Browser time", otherwise the selection box takes two lines.
- There is a vertical space issue, but if you rebase on main, it should be OK

nitzan-tz:
Let me work on it to add the missing parts

